Issue #16, opened 2020-03-06
Unexpected behaviour of alias analysis


CaviarChen:
**Program:**
position.wybe
```
pub type position { pub position(x:int, y:int) }

pub def printPosition(pos:position) use !io {
    !print(" (")
    !print(pos.x)
    !print(",")
    !print(pos.y)
    !println(")")
}
```
main.wybe
```
use position

?pos = position(100,150)
?pos2 = pos
!println("-------")
x(!pos, 200) 
!printPosition(pos)
!printPosition(pos2)
!println("-------")
x(!pos2, 90) 
!printPosition(pos)
!printPosition(pos2)
```

**Expected output:**
```
-------
 (200,150)
 (100,150)
-------
 (200,150)
 (90,150)
```

**Actual output:**
```
-------
 (200,150)
 (200,150)
-------
 (90,150)
 (90,150)
```

**Final dump:**
```
    foreign lpvm alloc(16:wybe.int, ?tmp$3#0:position.position)
    foreign lpvm mutate(~tmp$3#0:position.position, ?tmp$4#0:position.position, 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 100:wybe.int)
    foreign lpvm mutate(~tmp$4#0:position.position, ?tmp$0#0:position.position, 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 150:wybe.int)
    foreign c print_string("-------":wybe.string, ~#io#0:wybe.phantom, ?tmp$7#0:wybe.phantom) @wybe:100:39
    foreign c putchar('\n':wybe.char, ~tmp$7#0:wybe.phantom, ?#io#1:wybe.phantom) @wybe:86:26
    foreign lpvm mutate noalias(tmp$0#0:position.position, ?%pos#1:position.position, 0:wybe.int, 0:wybe.int, 16:wybe.int, 0:wybe.int, 200:wybe.int)
    position.printPosition<0>(pos#1:position.position, ~#io#1:wybe.phantom, ?#io#2:wybe.phantom) @alias_des2:7:2
    position.printPosition<0>(tmp$0#0:position.position, ~#io#2:wybe.phantom, ?#io#3:wybe.phantom) @alias_des2:8:2
    foreign c print_string("-------":wybe.string, ~#io#3:wybe.phantom, ?tmp$12#0:wybe.phantom) @wybe:100:39
[...]

PR #18, opened 2020-03-11
Fix LPVM mutate && Improve testing.

CaviarChen:
1. Fix issue #16 . (and update related `*.exp`)
2. Move current tests into a `final-dump` folder (and update related `*.exp` since the change of directory).
3. Add a new kind of testing `execution-test`.


CaviarChen:
The former part is contained in the commit: [Fix LPVM mutate](https://github.com/pschachte/wybe/pull/18/commits/51bb59b5db566637e980a9bfd4857e74da3677c3)
It just a single line of change in the `Blocks.hs`.

Issue #23, opened 2020-03-18
Unexpected behaviour of alias analysis 3


CaviarChen:
Code:
```
pub type intlist { pub [] | [|](head:int, tail:intlist) }

pub def list_print(x:intlist) use !io {
    if { x = [ ?h | ?t] :: 
        !print(h)
        !print(" ")
        !list_print(t)
    }
}

pub def list_println(x:intlist) use !io {
    !list_print(x)
    !println("")
}

?x = [1,2,3]
?y = [0]
!println("----------")
!list_println(x)
!list_println(y)

if { tail(!y, x) ::
    !println("----------")
    !print("x: ")
    !list_println(x)
    !print("y: ")
    !list_println(y)
}

if { head(!x, 999) ::
    !println("----------")
    !print("x: ")
    !list_println(x)
    !print("y: ")
    !list_println(y)
}

```

Output:
```
----------
1 2 3 
0 
----------
x: 1 2 3 
y: 0 1 2 3 
----------
x: 999 2 3 
y: 0 999 2 3
```

Expect Output:
```
----------
1 2 3 
0 
----------
x: 1 2 3 
y: 0 1 2 3 
----------
x: 999 2 3 
y: 0 1 2 3 
```

**The problem is that the output of mutate is always considered as unalised.**

PR #24, opened 2020-03-23
Fix issue #23

CaviarChen:
This PR is for #23 

The flag "noalias" on mutate instruction seems to be related. If the flag is set, then the input and output are considered unaliased after the mutate. 
https://github.com/pschachte/wybe/blob/cfce22d076d401c97389005bf308d2878e0341df/src/AliasAnalysis.hs#L198-L201

I don't understand the current condition for setting the "noalias" flag. My idea is that "noalias" should be set if and only if the data is a primitive value (not an address).


CaviarChen:
This fix is wrong, I'll create a new PR for the currect fix (to make the history looks clean).

Issue #25, opened 2020-03-23
The current implementation of Union-Find seems to be wrong.


CaviarChen:
I was trying to fix #21 which requires merging multiple alias maps and I found a case where the result is wrong.

`Util.combineUf (fromList [(a#0,a#0),(b#0,b#0),(r#0,a#0)]) (fromList [(a#0,a#0),(b#0,b#0),(r#0,b#0)])` returns `fromList [(a#0,a#0),(b#0,b#0),(r#0,a#0)]`

`Util.uniteUf (fromList [(a#0,a#0),(b#0,b#0),(r#0,b#0)]) (r#0) (a#0)`  returns `fromList [(a#0,a#0),(b#0,b#0),(r#0,a#0)]`.

I don't know how to fix it due to the current implementation doesn't make sense to me.
My questions are:
1. What is the "weight" for? https://github.com/pschachte/wybe/blob/f5cae83c5f2e6ee7163aba0987699c505ea74b2e/src/Util.hs#L162
2. If I understand it correctly, each node records its root directly instead of its parent.
https://github.com/pschachte/wybe/blob/f5cae83c5f2e6ee7163aba0987699c505ea74b2e/src/Util.hs#L273-L282
Each update requires visiting all nodes in the map and update old root to new root.
https://github.com/pschachte/wybe/blob/f5cae83c5f2e6ee7163aba0987699c505ea74b2e/src/Util.hs#L334-L339
I don't think it should be considered as union find since the union operation takes O(n) time instead of O(log n). Any naive approaches for this question run in O(n) (List of List, List of Set, Map from item to group id etc).

CaviarChen:
Oh, I got what the "weight" is for. It tries to update the "set" with less nodes in it? But still the whole thing doesn't make sense to me.

CaviarChen:
PR #26 replace the current `aliasMap` with a naive imeplementation and I thought about what to do next. 
In my opinion, I don't think we should use union find.
Note that, all the time complexities here (and also in my previous post) are missing a O(log n) factor beasue we are using immutable Map, Tree etc. But it doesn't change the conclusion.

In our use case, we need:
**Given two points, unify the set each point is in.**
Union find does handle this pretty well (in O(log n) with path compression). But I think in our case, _one of the points must be singleton due to the fact that we are using SSA_ (Plz help me verify this, there might be an exception when merging results from fork). If the conclusion is correst, then I think this step can easily be done in O(1).

**Remove one point**
[...]

Issue #21, opened 2020-03-15
Unexpected behaviour of alias analysis 2


CaviarChen:
Code:
```
pub type intlist { pub [] | [|](head:int, tail:intlist) }

pub def list_print(x:intlist) use !io {
    if { x = [ ?h | ?t] :: 
        !print(h)
        !print(" ")
        !list_print(t)
    }
}

pub def list_println(x:intlist) use !io {
    !list_print(x)
    !println("")
}

?x = [1,2,3,4,5,6]
?y = [999]
!list_println(x)
!list_println(y)

if { x = [ ?h | ?t] ::
    if { tail(!t, y) ::
        !print("t: ")
        !list_println(t)
    }
}

!list_println(x)
```

Ouput:
```
1 2 3 4 5 6 
999 
t: 2 999 
1 2 999 
```

Expect Output:
````
1 2 3 4 5 6 
999 
t: 2 999 
1 2 3 4 5 6 
````

The LPVM code is wrong:
```
*main* > public (0 calls)
0: (argc#0:wybe.int, [?argc#0:wybe.int], argv#0:wybe.int, [?argv#0:wybe.int], exit_code#0:wybe.int, [?exit_code#0:wybe.int], io#0:wybe.phantom, ?io#5:wybe.phantom): AliasPairs: []
    foreign lpvm alloc(16:wybe.int, ?tmp$13#0:error_case.intlist)
    foreign lpvm mutate(~tmp$13#0:error_case.intlist, ?tmp$14#0:error_case.intlist, 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 6:wybe.int)
    foreign lpvm mutate(~tmp$14#0:error_case.intlist, ?tmp$15#0:error_case.intlist, 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 0:error_case.intlist)
    foreign lpvm alloc(16:wybe.int, ?tmp$18#0:error_case.intlist)
    foreign lpvm mutate(~tmp$18#0:error_case.intlist, ?tmp$19#0:error_case.intlist, 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 5:wybe.int)
    foreign lpvm mutate(~tmp$19#0:error_case.intlist, ?tmp$20#0:error_case.intlist, 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~tmp$15#0:error_case.intlist)
[...]

PR #26, opened 2020-03-25
Fix Alias Analysis

CaviarChen:
This PR comtains fix for #21 and #25 .

For #21 , changes can be found here https://github.com/pschachte/wybe/commit/0a6393a984c32471588d593416266b62ba061d5f
For #25, this is only a quick and naive fix (replace it with a navie disjoint set), please don't close that issue.

### DIFF
--- a/src/Util.hs
+++ b/src/Util.hs
@@ -8,11 +8,11 @@
 {-# LANGUAGE DeriveGeneric #-}
 
 module Util (sameLength, maybeNth, checkMaybe, setMapInsert,
-             applyPair, fillLines, nop, sccElts,
-             UnionFind, UFInfo, unionFindToTransitivePairs,
-             initUnionFind, newUfItem, uniteUf, transformUfKey,
-             combineUf, removeFromUf, connectedToOthers,
-             removeDupTuples, pruneTuples, transitiveTuples, cartProd) where
+             applyPair, fillLines, nop, sccElts, DisjointSet, 
+             emptyDS, addOneToDS, unionTwoInDS, combineTwoDS, 
+             removeOneFromDS, removeFromDS, connectedToOthersInDS,
+             mapDS, filterDS, dsToTransitivePairs, removeDupTuples,
+             pruneTuples, transitiveTuples, cartProd) where
 
 
 import           Data.Graph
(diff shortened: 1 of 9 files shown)


src/Util.hs
```diff
@@ -151,9 +152,94 @@ cartProd ins outs = [(i, o) | i <- ins, o <- outs]
...
+    case _findFristInSet (Set.member x) ds of
+        Just _ -> ds
+        Nothing -> 
+            Set.insert (Set.singleton x) ds
+
+unionTwoInDS :: Ord a => a -> a -> DisjointSet a -> DisjointSet a
+unionTwoInDS x y ds = 
+    let xSet = _findFristInSet (Set.member x) ds in 
+    let ySet = _findFristInSet (Set.member y) ds in 
+        if (xSet == ySet) && (xSet /= Nothing)
```
pschachte: `(xSet == ySet) && (xSet /= Nothing)`  -- Swap order of those tests

src/Util.hs
```diff
@@ -151,9 +152,94 @@ cartProd ins outs = [(i, o) | i <- ins, o <- outs]
...
 ----------------------------------------------------------------
+-- TODO: Using Union-Find instead of this naive implementation
+-- The old implement is wrong https://github.com/pschachte/wybe/issues/25
+-- So we use this version as a quick fix.
+
+type DisjointSet a = Set (Set a)
+
+emptyDS = Set.empty
+
+_findFristInSet :: (a -> Bool) -> Set a -> Maybe a
```
pschachte: I suspect _findFristInSet should be _findFirstInSet.  But actually, _findOneInSet might be better, because foldr looks like it works from greatest to least element, so it's not clear what 'first' means.

pschachte:
Hi Zed,

Aside from the couple of comments I made, this looks OK.

CaviarChen:
> Hi Zed,
> 
> Aside from the couple of comments I made, this looks OK.

Hi Peter, 
I think I saw a comment about changing `_findFirstInSet` to `_findOneInSet`, but I can't find it now. Did you delete it? Do I still need to change it?

pschachte:
Yes, I started to write that comment in one place, then found I could insert it at the place in the code where it belongs, and then found the second one and deleted it, only to discover I’d deleted the wrong one.  I’m still finding github a bit confusing.  But yes, please do rename it.  I think it’s currently called _findFristInSet.

-Peter Schachte

PR #27, opened 2020-03-27
Fix #23 && a small improvement in AliasAnalysis

### LINKED ISSUE #23: Unexpected behaviour of alias analysis 3

CaviarChen:
This PR contains:
1. Fix for #23.
2. Remove the requirement that `member` has to be unaliased for destructive update.

I manually check the LPVM code in the expect test that chengd a lot, it should be correct.
I also added it as a execution test.


CaviarChen:
I removed the unused variable.

Issue #31, opened 2020-03-28
Support multiple specialisation


pschachte:
This allows generation of specialised versions of procedures based on how it will be called, and allows the compiler to select the best available version of a procedure for each call.

CaviarChen:
- [ ] refactor `SpeczVersionId ` and `AliasMultiSpeczInfo `. More detail can be found in the comments of https://github.com/pschachte/wybe/pull/35

pschachte:
Is it time to close this issue now (and open new ones for any unfinished aspects of this)?

CaviarChen:
Sure.

PR #33, opened 2020-04-02
Extend alias analysis to list interesting paramters of each procedure

CaviarChen:
This is part of #31


src/AliasAnalysis.hs
```diff
@@ -20,14 +20,16 @@ import           Data.List     as List
...
+import           Flow          ((|>))
 import           Options       (LogSelection (Analysis))
 import           Util
 
 
-----------------------------------------------------------------
---                 Proc Level Aliasing Analysis
-----------------------------------------------------------------
--- PROC LEVEL ALIAS ANALYSIS
+-- for local use only
```
pschachte, line 28: Please document this type

src/AliasAnalysis.hs
```diff
@@ -51,20 +53,31 @@ aliasSccBottomUp procs@(CyclicSCC multi) = do
...
     foldM (\info pspec -> do
         def <- getProcDef pspec
         let (ProcDefPrim _ _ analysis) = procImpln def
-        return $ info ++ [procArgAliasMap analysis]
+        return $ info ++ [extractAliasInfoFromAnalysis analysis]
         ) [] multi
 
 
+extractAliasInfoFromAnalysis :: ProcAnalysis 
+                                -> (AliasMap, AliasMultiSpeczInfo)
```
pschachte, line 71: Use AliasInfo type

src/AliasAnalysis.hs
```diff
@@ -51,20 +53,31 @@ aliasSccBottomUp procs@(CyclicSCC multi) = do
...
 currentAliasInfo procs@(CyclicSCC multi) =
     foldM (\info pspec -> do
         def <- getProcDef pspec
         let (ProcDefPrim _ _ analysis) = procImpln def
-        return $ info ++ [procArgAliasMap analysis]
+        return $ info ++ [extractAliasInfoFromAnalysis analysis]
         ) [] multi
 
 
+extractAliasInfoFromAnalysis :: ProcAnalysis
```
pschachte, line 70: Document this function

test-cases/final-dump/type_list.exp
```diff
@@ -46,7 +48,9 @@ AFTER EVERYTHING:
 
 
 ++ > public (2 calls)
-0: ++(x#0:type_list.intlist, y#0:type_list.intlist, ?$#0:type_list.intlist): AliasPairs: [($#0,y#0)]
+0: ++(x#0:type_list.intlist, y#0:type_list.intlist, ?$#0:type_list.intlist):
+ AliasPairs: [($#0,y#0)]
```
pschachte, line 52: Please check this.  I would have expected $ to be aliased to both x and y.

src/AliasAnalysis.hs
CaviarChen: Added.

src/AliasAnalysis.hs
CaviarChen: Added.

src/AliasAnalysis.hs
CaviarChen: Actually, `AliasInfo` is `(AliasMap, AliasMultiSpeczInfoLocal)` and here we need `(AliasMap, AliasMultiSpeczInfo)`.

Both `AliasInfo` and `AliasMultiSpeczInfoLocal` are only used in this module. I am not sure if the design is good but my idea is that the local one is used for the analysis (storing temp things and current state) and the non-local one is the actual one that will be stored in the LPVM module (describing the overall info about a procedure).

test-cases/final-dump/type_list.exp
CaviarChen: I think the current result is correct. Becasue the whole content from varibale `x` will be reconstructed. The return value only contains the original `y`.

For your reference, here is the code and lpvm instruction:
```
pub def ++(x:intlist, y:intlist):intlist =
    if x = [ ?h | ?t]
    then [h|t++y]
    else y
```

```
++ > public (2 calls)
0: ++(x#0:type_list.intlist, y#0:type_list.intlist, ?$#0:type_list.intlist):
 AliasPairs: [($#0,y#0)]
 AliasMultiSpeczInfo: []
    foreign llvm icmp ne(x#0:type_list.intlist, 0:wybe.int, ?tmp$5#0:wybe.bool)
    case ~tmp$5#0:wybe.bool of
    0:
        foreign llvm move(~y#0:type_list.intlist, ?$#0:type_list.intlist) @type_list:3:5

    1:
        foreign lpvm access(x#0:type_list.intlist, 0:wybe.int, ?h#0:wybe.int)
        foreign lpvm access(~x#0:type_list.intlist, 8:wybe.int, ?t#0:type_list.intlist)
        type_list.++<0>(~t#0:type_list.intlist, ~y#0:type_list.intlist, ?tmp$2#0:type_list.intlist) @type_list:5:13
        foreign lpvm alloc(16:wybe.int, ?tmp$8#0:type_list.intlist)
        foreign lpvm mutate(~tmp$8#0:type_list.intlist, ?tmp$9#0:type_list.intlist, 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~h#0:wybe.int)
        foreign lpvm mutate(~tmp$9#0:type_list.intlist, ?$#0:type_list.intlist, 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~tmp$2#0:type_list.intlist)
```

src/AliasAnalysis.hs
pschachte: I don't think that the fact these things are only used in this module is a good reason not to document them.  At some point, someone is going to need to understand them to maintain the code.  I'm not asking you to document Ting's code (I should have asked her to do that), but please do document the new code and types you're putting in.

test-cases/final-dump/type_list.exp
pschachte: Yup, you're right.  I overlooked the fact that it's an int list.

src/AliasAnalysis.hs
```diff
@@ -25,7 +25,10 @@ import           Options       (LogSelection (Analysis))
 import           Util
 
 
--- for local use only
+-- This type is for local use only. It stores interesting
+-- parameters during the analysis. After the analysis, it
+-- will be converted to [AliasMultiSpeczInfo] and stored
+-- in LPVM modure.
```
pschachte, line 31: Can you briefly explain what "interesting" means?

src/AliasAnalysis.hs
CaviarChen: Sure, I wasn't trying to argue about the documentation in my comment above. I am just saying that the `AliasInfo` is not the same as `(AliasMap, AliasMultiSpeczInfo)`.

pschachte:
Yeah, I didn’t see your change to the doc when I wrote that.  Then I found the change.

I’m still finding github a bit hard to use.

-Peter Schachte

PR #35, opened 2020-04-10
Replace "alloc" with "move" by reusing dead memory cells if possible.

CaviarChen:
This is part of #31.

The current aliasMap is undirectional, so we use this as a little "hack" to achieve destructive list reverse.


src/AliasAnalysis.hs
```diff
@@ -450,6 +465,111 @@ isArgVarInteresting _aliasMap _ = False
...
+
+
+-- Find new dead cell from the given "primArgs" of "access" instruction.
+updateDeadCellsByAccessArgs :: [PrimVarName] -> (AliasMap, DeadCells) 
+        -> [PrimArg] -> Compiler DeadCells
+updateDeadCellsByAccessArgs aliasedParams (aliasMap, deadCells) primArgs = do
+    -- [struct:type, offset:int, ?member:type2]
+    let [struct, _, _] = primArgs
+    let ArgVar{argVarName=varName, argVarType=ty, argVarFinal=final} = struct
+    rep <- lookupTypeRepresentation ty
```
CaviarChen, line 533: @pschachte How heavy is the `lookupTypeRepresentation`?
The code seems to be a bit complicated and we are using it a lot here as well as when updating `aliasMap`.

src/AST.hs
```diff
@@ -1583,19 +1585,32 @@ showSuperProc (SuperprocIs super) =
...
                                          -- ^defn in LPVM (clausal) form
       -- defn in SSA (LLVM) form along with any needed extern definitions
     deriving (Eq,Generic)
 
 
+-- | The identity of each specialized version. It needs to be a bijection
+-- between this Id and the specialized condition. 
+-- Currently, [Int] is enought. We can change it to [Tuple] when there 
+-- are more things to consider. More detial can be found in [Transform.hs]
+type SpeczVersionId = Int
```
pschachte: A bijection is too much to ask for.  Some specialisations will be for things like specific values (eg, specialising a higher order function for the higher order argument).

src/AST.hs
```diff
@@ -1606,19 +1621,27 @@ showAliasMap aliasMap = show $ aliasMapToAliasPairs aliasMap
 aliasMapToAliasPairs :: AliasMap -> [(PrimVarName, PrimVarName)]
 aliasMapToAliasPairs aliasMap = Set.toList $ dsToTransitivePairs aliasMap
 
+-- | Multiple specialization info for global alias
```
pschachte: This should be generalised to be a set of specialisation elements (or some such name), where specialisation element is a type that we will extend over time.  For now, one specialisation element can just be something like (KnownUnique Int) where the int is an argument number.

CaviarChen: Will resolve this later. https://github.com/pschachte/wybe/issues/31

src/AST.hs
CaviarChen: Will resolve this later. https://github.com/pschachte/wybe/issues/31

src/Transform.hs
```diff
@@ -112,29 +121,57 @@ transformForks caller body aliasMap aliasedParams = do
...
+                    <- updateDeadCellsByAccessArgs aliasedParams 
+                                        (aliasMap, deadCells) args
+                return (primc, deadCells')
+            PrimForeign "lpvm" "alloc" _ args  -> do
+                let (result, deadCells') =
+                        assignDeadCellsByAllocArgs aliasedParams deadCells args
+                let primc' = case result of 
+                        Nothing -> primc
+                        Just (selectedCell, _) -> 
+                            -- replace "access" with "move" by reusing the
```
pschachte, line 153: I think "access" -> "alloc"

src/Transform.hs
```diff
@@ -112,29 +121,57 @@ transformForks caller body aliasMap aliasedParams = do
...
+                        Just (selectedCell, _) -> 
+                            -- replace "access" with "move" by reusing the 
+                            -- "selectedCell".
+                            let [_, varOut] = args in
+                            let varIn = 
+                                    varOut {argVarName = selectedCell,
+                                            argVarFlow = FlowIn,
+                                            argVarFinal = True}
+                            in
+                            PrimForeign "llvm" "move" [] [varIn, varOut]
```
pschachte, line 161: After doing this, we probably want to re-run the optimiser.  Often some of the fields being stored in the newly allocated memory are actually already stored in the structure we're now reusing, and it would be good to avoid redundant instructions.  In fact, it would be worth trying the alias analysis and transformation before the optimiser; if it works well enough on un-optimised code, that would allow the optimiser to improve code after it.

src/AliasAnalysis.hs
```diff
@@ -450,6 +465,111 @@ isArgVarInteresting _aliasMap _ = False
...
+-- The currently alias map is undirectional, so we have to identify
+-- dead cells using access instruction. 
+-- if there is an access instruction that reads some value from x,
+-- then we consider x is dead if x is unaliased and final.
+--
+-- The transform part can be found in "Tranform.hs".
+
+-- TODO: reuse cells that are different types but has the same size.
+-- TODO: call "GC_free" on large unused dead cells.
+
```
pschachte, line 488: Another TODO:  we'd like this analysis to detect structures that are dead aside from later access instructions, which could be moved earlier to allow the structure to be reused.

test-cases/execution/multi_specz_reverse.wybe
```diff
@@ -0,0 +1,37 @@
+pub type intlist { pub [] | [|](head:int, tail:intlist) }
+
+pub def list_print(x:intlist) use !io {
+    if { x = [ ?h | ?t] :: 
+        !print(h)
+        !print(" ")
+        !list_print(t)
+    }
+}
+
```
pschachte, line 10: why not just:
```
pub def list_reverse(lst:intlist):intlist = list_reverse_helper(lst, [])

pub def list_reverse_helper(lst:intlist, acc:intlist):intlist =
    if lst = [?h | ?t] then list_reverse_helper(t, [h|acc]) else acc
```
(untested, but should be close to correct).

test-cases/final-dump/alias4.exp
```diff
@@ -56,7 +56,7 @@ replicate1 > public (1 calls)
     foreign c print_int(~tmp$1#0:wybe.int, ~#io#1:wybe.phantom, ?tmp$14#0:wybe.phantom) @wybe:nn:nn
     foreign c putchar('\n':wybe.char, ~tmp$14#0:wybe.phantom, ?#io#2:wybe.phantom) @wybe:nn:nn
     foreign lpvm access(~p1#0:position.position, 0:wybe.int, ?tmp$3#0:wybe.int)
-    foreign lpvm alloc(16:wybe.int, ?tmp$18#0:position.position)
+    foreign llvm move(~tmp$0#0:position.position, ?tmp$18#0:position.position)
```
pschachte, line 59: Cool!

src/Transform.hs
CaviarChen: Sure, but I think the work of this task is big enough for a separate PR. Added it as a TODO.

pschachte: Oh, absolutely.  This is a longer term issue.

PR #39, opened 2020-04-18
Support reading int float and string.

CaviarChen:


wybelibs/cbits.c
```diff
@@ -10,12 +10,12 @@
 #include <math.h>
 
 // putchard - putchar that takes a double and returns 0.
```
pschachte, line 12: That comment looks wrong

CaviarChen:
It actually depends on you. We can have an agreement that I am allow to merge the PR only if you approved it and all comments are resolved. 
Currently, there isn't any branch protection rules so the "Squash and merge" is button is green even if you haven't approved it or the CI fails.

PR #45, opened 2020-04-24
Implement a better way of handling parameters in alias analysis.

CaviarChen:
This PR is a part of #31.

This PR introduces a better way of handling parameters in alias analysis, so we can have a preciser analysis and more optimize opportunities.

In this approach, we consider a parameter as two things, a normal variable and a thing that the variable is (or maybe) alias to outside the procedure scpoe. More detail can be found at around line 37 in `AST.hs`.

### DIFF
--- a/src/AliasAnalysis.hs
+++ b/src/AliasAnalysis.hs
@@ -167,7 +167,8 @@ aliasProcDef def = return def
 
 
 -- Analysis a "ProcBody"
--- It doesn't return "DeadCells".
+-- It returns "AliasMap" and "AliasMultiSpeczInfoLocal" but not "DeadCells"
+-- since we don't need to merge it and have the result after the analysis.
 aliasedByBody :: PrimProto -> ProcBody -> AnalysisInfo 
         -> Compiler (AliasMapLocal, AliasMultiSpeczInfoLocal)
 aliasedByBody caller body analysisInfo =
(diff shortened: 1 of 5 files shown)


src/AliasAnalysis.hs
```diff
@@ -398,51 +416,53 @@ mergeMultiSpeczInfo = List.foldl Set.union Set.empty
...
             return multiSpeczInfo'
         PrimForeign "lpvm" "mutate" flags args -> do
-            let [fIn, _, _, _, _, _, _] = args
-            let ArgVar{argVarName=inName} = fIn
-            if List.elem inName realParams 
-                && isArgVarInteresting aliasMap fIn
-            then do 
-                logAlias $ "Found interesting param: " ++ show inName
-                return $ Set.insert inName multiSpeczInfo
+            let [fIn, _, _, ArgInt des _, _, _, _] = args
```
pschachte, line 454: More robust to use a case rather than a let, and have a shouldnt for cases that don't match.  Or pattern match in the LHS of the equation, and have another equation for non-matching lpvm mutates.

src/AliasAnalysis.hs
```diff
@@ -398,51 +416,53 @@ mergeMultiSpeczInfo = List.foldl Set.union Set.empty
...
+            -- Skip "mutate" that is set to be destructive by previous optimizer
+            if des /= 1
```
pschachte, line 456: Better still, have two equations for mutate, one for ArgInt 0 for the destructive flag and one for ArgInt 1.  Longer term, we'll want to support having a Boolean variable for the destructive flag, but for now have a shouldnt for the case of a destructive flag that's not a static constant.

src/AliasAnalysis.hs
```diff
@@ -140,10 +166,10 @@ aliasProcDef def
 aliasProcDef def = return def
 
 
--- Analysis a [ProcBody]
--- It doesn't return [DeadCells].
+-- Analysis a "ProcBody"
+-- It doesn't return "DeadCells".
```
pschachte, line 170: It would be more useful to say what it *does* return than what it *doesn't*.

Issue #51, opened 2020-05-14
When compiling a statement like foo(?x, ..., x) the wrong '#version' suffix is added to the x


pschachte:

pschachte:
Fixed by #55.

PR #55, opened 2020-05-14
 Maintain two maps while numbering variables, one for variable assign…

pschachte:
…ments and one for uses.  At the end of each statement, the current variable one becomes the assignments one.  This ensures that variable uses get the right number even when they appear after an assignment in the statement.


pschachte:
Closes #51

PR #61, opened 2020-06-05
A temporary of final marking && Fix tmp count returned by Bodybuilder

CaviarChen:


src/Optimise.hs
```diff
@@ -86,9 +86,16 @@ optimiseProcDefBU pspec def = do
     logOptimise $ "*** " ++ show pspec ++
       " before optimisation:" ++ showProcDef 4 def
     def' <- procExpansion pspec def >>= decideInlining
+    -- XXX The second run of "procExpansion" is just a temporary fix, we need
+    -- something better.
+    -- The actual issue is in the BodyBuilder, some optimizations after the
+    -- "final marking" can make the final flag becomre invalid.
```
pschachte, line 92: It's not actually invalid, it just fails to flag a last use of some variables.

Issue #59, opened 2020-06-01
Wybe program runs slower on system with more cores


CaviarChen:
This is a bit odd. The same program can become slower when there are more cpu cores in the system.
I used the same physical machine to run vm and allocate a different number of cpu cores to it. Then I ran the same wybe program on it and here are the results:
| number of cores | time used |
|-----------------|-----------|
| 1               | 54s       |
| 2               | 58s       |
| 6               | 128s      |
| 12              | 220s      |

I suspect that the issue is caused by the gc, because:

1. The situation is worse on a program with more memory allocation.
2. Wybe program will create multiple threads with the number equal to the number of cores in the system.

After a quick look at the doc of the BDW GC, it seems that the gc should be single-threaded with incremental mode disabled by default. Not sure why and I will keep investigating this.

pschachte:
Interesting.  Can you try a compute-bound process and see if that behaves differently?  Could it be a VM issue?

CaviarChen:
> Interesting. Can you try a compute-bound process and see if that behaves differently? Could it be a VM issue?

I don't think it's a VM issue. It's very hard to find machines that are identical expect number of cores. So I tried different virtualization technologies and the issue still exists. I also try to run it on my friend's macbook which is 8 cores (16 logical cores) and it's about 3x slower than mine (2 cores and 4 logical cores, and my cpu is a lot older than my friend's). 

I didn't find a way to hide cores from a program.
I tried to use `taskset` to force a program to run on a given core only. 
Since it still know the number of cores the system has, so it starts that many threads as usual. It still runs slower than the 1-core case, but it's faster than run it without `taskset` (I think that's odd since now it's 12 threads squashed into 1 core)

I also tried to use `echo 0 | sudo tee /sys/devices/system/cpu/cpu1/online` to disable other cores and the results are the same as the previous one.

Removing the bdw gc did solve this issue.

CaviarChen:
The parallel marking of bdw gc is definitely the cause. The issue can be avoided by setting the environment variable `GC_NPROCS=1`. 

However, I am not sure how to fix this. I couldn't find any api to set the `GC_NPROCS` in runtime except this environment variable. And building a single-threaded bdw-gc seems to be a bit annoying to fit with our current makefile.

Here are some doc related to this for you reference: (By the way, I think the doc of bdw gc is a bit confusing)

[...]

PR #62, opened 2020-06-09
Mitigate performance issue in GC (workaround) && Prevent optimiser to remove the call to gc_init

CaviarChen:
More detail can be found here: #59

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -856,9 +856,10 @@ buildMain mainImports =
                     | m <- mainImports]
         -- XXX Shouldn't have to hard code assignment of phantom to io
         -- XXX Should insert assignments of initialised visible resources
-        bodyCode = [move (castTo (iVal 0) phantomType) (varSet "io"),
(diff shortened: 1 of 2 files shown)


src/Builder.hs
```diff
@@ -856,9 +856,10 @@ buildMain mainImports =
                     | m <- mainImports]
         -- XXX Shouldn't have to hard code assignment of phantom to io
         -- XXX Should insert assignments of initialised visible resources
-        bodyCode = [move (castTo (iVal 0) phantomType) (varSet "io"),
-                    move (intCast $ iVal 0) (intVarSet "exit_code"),
-                    Unplaced $ ForeignCall "c" "gc_init" [] []] ++ bodyInner
+        bodyCode = [move (intCast $ iVal 0) (intVarSet "exit_code"),
```
pschachte, line 859: I think you should keep the move (castTo (iVal 0) phantomType) (varSet "io") instruction, and change the (varSet "io") argument to gc_init to (varGetSet "io"), so it threads though the io state, rather than just generating it.

Issue #60, opened 2020-06-03
Linktime dead code elimination only works on macOS


CaviarChen:
We added the support of linktime dead code elimination in #4 when wybe only works on macOS.
It turns out that the `-dead_strip` only exists in the system linker on macOS so it doesn't work on ubuntu.

I tried `-Wl,--gc-sections` and it doesn't work, I believe it requires that every function in its own ELF section.

CaviarChen:
I took a further look into this issue and noticed the following things:

In terms of multiple specialization, we don't really need link-time dead code elimination.
Currently, multiple specialization only generates versions that we need. The missing specializations of a proc will be added into the module during the top-down pass, and we accept a module that has more specialization versions than we need (a superset). We can remove those during the top-down pass if we really don't want them and don't want to use link-time dead code elimination. That's super easy to do in the current top-down pass.

In terms of making the final executable smaller, the main factor is the lpvm section. The final executable contains a concatenation of all lpvm section of all object files it linked. This section usually takes 90% of the size of the executable and it is totally useless.


Back to the link-time dead code elimination.

The system linker on macOS provides an arg `-dead_strip` which can remove unused code and sections (such as the lpvm section) without any requirements.

On Linux, there isn't an equivalent arg. There is `-Wl,--gc-sections` which can remove unused code. However, it can only remove code by section, so we have to put procedures into different sections instead of a single section. For procedure foo, it should be in `.text.foo` instead of `.text`. This may make the object code a little bit bigger since there are more sections now. I tested this and it's doable.
Another problem is that `-Wl,--gc-sections` can't remove the lpvm section. I don't know why though.
To remove the lpvm section on Linux, I found the following two approaches:
1. Use cli `objcopy` to remove that section after we got the executable. It's simple but the problem is that the linker still needs to gather all lpvm sections and write to the executable file first.
2. Use a custom version of the linker script to discard this section, so the linker can skip it directly. It's nice but I am not sure what is the best practice of using a custom linker script. It seems a bit complicated.

Overall, I am not sure what approaches should we use and how we handle the difference between those two platforms. For example, If we are going to use `-Wl,--gc-sections` on Linux, then should we put procedures in different sections on macOS?

CaviarChen:
@pschachte In our last meeting, we decided to put each function in a separate section for `-Wl,--gc-sections` on Linux and keep the same behavior on macOS. However, I just realized that the section specifier of ELF and mach-o are different.
[...]

PR #70, opened 2020-07-11
Supporting link-time dead code elimination on Linux

CaviarChen:
Related issue: #60 

Note that I didn't put all platform-specific things in `Config.hs`, because I don't think they fit the context of config. You can find all of them (including previous ones) by searching usages of `buildOS`.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -10,13 +10,15 @@
 --  OSes.
 module Config (sourceExtension, objectExtension, executableExtension,
                interfaceExtension, bitcodeExtension,
-               ldArgs, wordSize, wordSizeBytes,
+               wordSize, wordSizeBytes,
                availableTagBits, tagMask, smallestAllocatedAddress,
-               assemblyExtension, archiveExtension, magicVersion)
+               assemblyExtension, archiveExtension, magicVersion,
+               linkerDeadStripArgs, functionDefSection)
     where
 
 import Data.Word
 import qualified Data.List as List
+import Distribution.System (buildOS, OS (..))
 import Foreign.Storable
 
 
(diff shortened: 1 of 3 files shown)


src/Config.hs
```diff
@@ -92,12 +94,29 @@ sharedLibDirName :: String
...
 magicVersion =
     let magicStr = "WB01"
     in List.map (fromIntegral . fromEnum) magicStr
+
+
+-- | Arguments of dead code elimination for linker
+-- Look for "Link time dead code elimination" in "Emit.hs" for more detail
+linkerDeadStripArgs :: [String]
+linkerDeadStripArgs =
+    case buildOS of
```
pschachte, line 108: I'd like to keep platform dependencies in Config.hs

src/Config.hs
```diff
@@ -92,12 +94,29 @@ sharedLibDirName :: String
...
+        Linux -> ["-Wl,--gc-sections"]
+        _     -> error "Unsupported operation system"
+
+
+-- | Given the name of a function and return which section to store it,
+-- Nothing means the default section. For now, we put functions in separate
+-- sections on Linux to fit the "-Wl,--gc-sections" above.
+functionDefSection :: String -> Maybe String
+functionDefSection label =
+    case buildOS of
```
pschachte, line 119: I'd like to keep platform dependencies in Config.hs

src/Emit.hs
```diff
@@ -228,26 +237,44 @@ suppressLdWarnings s = intercalate "\n" $ List.filter notWarning $ lines s
     notWarning l = not ("ld: warning:" `List.isPrefixOf` l)
 
 
+-- | Remove the lpvm section from the given file. It's only effective on Linux,
+-- since we don't need it on macOS.
+removeLPVMSection :: FilePath -> IO (Either String ())
+removeLPVMSection target =
+    case buildOS of
```
pschachte, line 244: I'd like to keep platform dependencies in Config.hs

src/Config.hs
CaviarChen: It's already in Config.hs.

src/Config.hs
CaviarChen: It's already in Config.hs.

src/Emit.hs
CaviarChen: Moved.

CaviarChen:
@pschachte All platform-specific things in this PR are moved to `Config.hs`, but there are some old code that are platform-specific and are outside `Config.hs`.  Mainly in `ObjectInterface.hs` and `Macho.hs`.

Issue #66, opened 2020-06-17
Incremental build may generate object file with outdated code


CaviarChen:
main.wybe
```
use lib

!foo()
```

lib.wybe
```
pub def foo() use !io {
    !println("test message A")
}
```

Reproduce steps:

1. Build `main`.
2. Change `foo()` in `lib.wybe` (eg. change the string in `println`).
3. Build `main` again and this new main will behave the same as the previous one.

The current incremental build works as follows:
https://github.com/pschachte/wybe/blob/5cb70a87eb1af7738907d163469d19773a74fbfb/src/Builder.hs#L327-L343
During the bottom-up pass, if the object file is newer than the source code file (based on last modified time) then the compiler will re-use the lpvm module read from the object file.

https://github.com/pschachte/wybe/blob/5cb70a87eb1af7738907d163469d19773a74fbfb/src/Builder.hs#L949-L959
After the top-down pass, the same logic is used to determine whether the object file needs to be updated (It is disabled by PR #50 ).

The main issue is that incremental build only considers each module itself and ignores the changes of its dependencies. Inlining and alias analysis that cross multiple modules will break.

Supporting incremental build during the top-down pass requires this issue to be fixed first. I'm planning to discuss this in the coming meeting.

Issue #54, opened 2020-05-14
Compiler trys to inline uncompiled code.


CaviarChen:
There is a case where https://github.com/pschachte/wybe/blob/47a01b25d13271fb96a3c6e814603ccb51397f2c/src/Expansion.hs#L198-L201 will be reached with `procImpln def` is `ProcDefSrc` instead of `ProcDefPrim`.

A sample csae is `test-cases/final-dump/multi_specz_cyclic_lib2.wybe`.
Actually it belongs to a test case with 3 files.
```
test-cases/final-dump/multi_specz_cyclic_exe.wybe
test-cases/final-dump/multi_specz_cyclic_lib.wybe
test-cases/final-dump/multi_specz_cyclic_lib2.wybe
```
Compiling them as a whole succeeds without any error but compiling `multi_specz_cyclic_lib2.o` individually causes this problem.

CaviarChen:
I believe this issue is caused by #68.

PR #71, opened 2020-07-15
Update builder to fix #68

### LINKED ISSUE #68: Incremental build may cause the builder to call compileModSCC with invalid SCC.
The diagram below shows the target main and its dependencies.
dep

The problem is related to the circular dependency of `ca` and `cb`.
If there isn't any object file that can be reused, then `compileModSCC [ca, cb]` will be executed, which is correct.
If there are up-to-date object files for `ca` and `cb`, then `compileModSCC` won't be executed, which is correct.
If only the object file of `ca` is missing (or outdated), then `compileModSCC [ca]` will be executed, which is wrong.
if only the object file of `cb` is missing (or outdated), then `compileModSCC` won't be executed and left `cb` uncompiled, which is wrong.

The cause is that `loadModuleFromObjFile` inserts modules directly into the Compiler State and doesn't call things like `exitModule`, so the topological sort is kind of broken. In the example above, `cb` is deferred first. Then compiler visits `ca` and realize that `ca` and `cb` is in an SCC. But if `cb` is loaded directly then it won't be marked as deferred and if `ca` is loaded directly then the deferred `cb` won't be picked.

Overall, I feel that the builder part is a bit messy and we need to fix this before working on #66 
In my opinion, a module has two stages while building (shown in the diagram below). 

Currently, the builder handles the first part correctly. A `stage 1 module` only depends on its source code, so using the lastModifiedTime is correct.
However, a `stage 2 module` (generated by the `compileModSCC`) also depends on other `stage 2 modules` and its `stage 1 module`, so we can't let the decision from the first part affects this part. So we need to visit all SCCs no matter if it's loaded from an object file or not. And for each SCC, we need to check whether anything changes (like the one that will be added to fix #66 ), if so, then `compileModSCC` should be executed with the whole SCC.
Then there is another problem, currently, we only store the `stage 2 modules` in the object file, we need a way to get the `stage 1 module` if we need to rebuild it. I don't like the idea of getting it from the source code since it basically makes the object file not portable at all. Maybe storing the list of tokens?

CaviarChen:
This PR is a bug fix for #68 and a preparation for #66.

Our current testing system doesn't support incremental build. I tested some cases manualy. Hoever, there is a bug that stops me for some cases, and I believe it is related to #54 

Some approches in this are a bit hacky, I left some XXXs in it.

The error case in #66 is handled like this:
```
Builder:  === Start running bottom-up pass
Builder:  --- Running on: wybe, wybe.string, wybe.phantom, wybe.io, wybe.int, wybe.float, wybe.command_line, wybe.char, wybe.bool
Builder:  Skip this SCC
Builder:  --- Running on: down
Builder:  Skip this SCC
Builder:  --- Running on: ca, cb, ca.intlist
Builder: Mods that need to be reloaded: ca, ca.intlist
Builder: Root mods: fromList [((["ca"],"/home/ubuntu/Documents/wybe/workspace/complex_dep/ca.wybe"),[["ca","intlist"],["ca"]])]
Builder: Reload root mod: ca contains: ca.intlist, ca from: "/home/ubuntu/Documents/wybe/workspace/complex_dep/ca.wybe"
Builder: ===> Compiling module ca
Builder: HASH: b6d0bfc0a0df4fe0e0ecf6cbbc4faf17bc392992
Builder: building dependencies: ca.intlist, cb, wybe
Builder: Load dependency: ca.intlist
Builder: Building ca.intlist with library directories /home/ubuntu/Documents/wybe/workspace/complex_dep, /usr/local/lib/wybe
Builder: module ca.intlist is already loaded
Builder: Load dependency: cb
Builder: Building cb with library directories /home/ubuntu/Documents/wybe/workspace/complex_dep, /usr/local/lib/wybe
Builder: module cb is already loaded
Builder: Load dependency: wybe
Builder: Building wybe with library directories /home/ubuntu/Documents/wybe/workspace/complex_dep, /usr/local/lib/wybe
Builder: module wybe is already loaded
Builder: <=== finished compling module ca
Builder:      module dependency SCC: ca, ca.intlist
Builder: ~=~ [["ca"],["ca","intlist"],["cb"],["down"],["main"],["top"],["wybe"],["wybe","bool"],["wybe","char"],["wybe","command_line"],["wybe","float"],["wybe","int"],["wybe","io"],["wybe","phantom"],["wybe","string"]]
Builder: compileModSCC: [ca, cb, ca.intlist]

```


src/Builder.hs
```diff
@@ -366,6 +372,103 @@ buildModuleIfNeeded force modspec possDirs = do
                     return False
 
                 _ -> shouldnt "inconsistent file existence"
+    -- Actual building
+    -- XXX We should really consider refactoring the whole thing such that it
+    -- has 3 explicit step. 1. loads all module (source code). 2. compute the
```
pschachte, line 377: It shouldn't be necessary to load source code for unchanged files, because the object file should have everything that is needed.

src/Builder.hs
```diff
@@ -366,6 +372,103 @@ buildModuleIfNeeded force modspec possDirs = do
...
+    -- 1. "compileModSCC" needs to be separated from loading anyway since we
+    --    need to reload a module from source if it's dependencies changed. 
+    --    And since currently the top-sort is combined with loading, reload a
+    --    module will mess up a lot of things. And for the same reason,
+    --    "buildModuleIfNeeded" with topLevel = True can only be called once.
+    -- 2. Modules have things like "minDependencyNum" which we don't really
+    --    need to store in module.
+    -- 3. Similar things need to be handled twice for loading from source code
+    --    and object file. That is pretty error-prone.
+    when topLevel $ do
```
pschachte, line 389: There are really two different issues here:

  1.  Disentangling the SCC algorithm from module loading; and
  2.  Disentangling the processing of module SCCs from loading.

Item (1) just doesn't seem like a big deal to me either way, except that it can't be done unless (2) is done.  The reason these are entangled is that I wanted to be able to compile modules bottom-up by SCC, so that only object code and analysis results would need to be kept around while compiling later modules (ie, to minimise memory usage during compilation).

We should discuss this and work out which way to go on this.

PR #75, opened 2020-07-24
Fix #66 by using the hash of module's public interface.

### LINKED ISSUE #66: Incremental build may generate object file with outdated code

CaviarChen:
#66

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -753,15 +753,16 @@ compileModSCC mspecs = do
 
 
 
-
+updateInterfaceHash :: ModSpec -> Compiler ()
 updateInterfaceHash mspec = do
     reenterModule mspec
-    interface <- getModule modInterface
+    interface <- getModuleInterface
     let hash = hashInterface interface
     updateModule (\m -> m {modInterfaceHash = hash})
     reexitModule 
 
 
+updateImportsInterfaceHash :: ModSpec -> Compiler ()
 updateImportsInterfaceHash mspec = do
     reenterModule mspec
     updateModImplementationM (\imp -> do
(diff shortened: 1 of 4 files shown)


src/Builder.hs
```diff
@@ -563,12 +563,39 @@ compileModBottomUpPass orderedSCCs = do
...
+--      the current interface hash.
 isCompileNeeded :: [ModSpec] -> Compiler Bool
 isCompileNeeded modSCC = do
     unchanged <- gets unchangedMods
-    return $ not $ List.all (`Set.member` unchanged) modSCC
+    if List.all (`Set.member` unchanged) modSCC
+    then do
+        upToDate <- List.and <$> mapM (\m -> do
+            imports <- getModuleImplementationField modImports `inModule` m 
+            List.and <$> mapM (\(m', (_, hash)) ->
```
pschachte, line 577: You could use [andM](https://hackage.haskell.org/package/extra-1.7.4/docs/Control-Monad-Extra.html#v:andM) from Control.Monad.Extra.

PR #79, opened 2020-08-01
Add CallSiteID && use CallSiteID in MultiSpeczDepInfo

CaviarChen:
We discussed it [here](https://github.com/pschachte/wybe/pull/64#discussion_r444640251).


CaviarChen:
I disabled the CI on macOS due to a problem with caching, the ubuntu one should be enough from most cases. Will fix it in a later PR.

src/AST.hs
```diff
@@ -1652,12 +1652,8 @@ aliasMapToAliasPairs aliasMap = Set.toList $ dsToTransitivePairs aliasMap
 -- For a given specialization version of the current proc, this info should be
 -- enough to compute all specz versions it required. A sample case is
 -- "expandSpeczVersionsAlias" in "Transform.hs".
--- XXX this map is per call site, so the args is included to separate different
--- calls to the same proc. The current approach is not currect, we should give
--- each call site a unique id similar to the tmep variable count in
--- "BodyBuilder" and use it as the key.
 type MultiSpeczDepInfo = 
-        Map (ProcSpec, [PrimArg]) (Set CallSiteProperty)
+        Map CallSiteID (ProcSpec, Set CallSiteProperty)
```
pschachte, line 1656: Please document what the ProcSpec here is (caller or callee).

src/AST.hs
```diff
@@ -2227,14 +2223,20 @@ data PrimVarName =
 -- |A primitive statment, including those that can only appear in a
 --  loop.
 data Prim
-     = PrimCall ProcSpec [PrimArg]
+     = PrimCall (Maybe CallSiteID) ProcSpec [PrimArg]
```
pschachte, line 2226: I think it would be better to  assign the CallSiteID in Clauses.hs (in compileSimpleStmt').  It should be easy enough to add a counter to the ClauseCompState.  Then later code would just need to preserve it when it transforms calls.  That would allow PrimCall to hold a CallSiteID rather than a Maybe CallSiteID.

src/Expansion.hs
```diff
@@ -127,12 +129,26 @@ addRenaming var val = do
...
+            inliningNow <- gets inlining
+            if inliningNow
+            then do
+                callSiteID <- gets nextCallSiteID
+                modify (\st -> st {nextCallSiteID = callSiteID + 1})
+                return $ PrimCall callSiteID pspec args
+            else 
+                return prim
+        _ -> return prim
+    lift $ instr prim pos
```
pschachte, line 145: Shouldn't that be prim' there?

CaviarChen: Right! I'll update the call_site_id case to test the inlining.

PR #84, opened 2020-08-11
Make "Dead Memory Cell Analysis" aware of the size of memory cells.

CaviarChen:


documentation/lpvm.md
```diff
@@ -7,8 +7,10 @@ the LLVM instructions.
    Allocate (at least) *size* bytes of memory and store the address in
    *struct*, which has its specified type.
 
-* `foreign lpvm access(`*struct:type*, *offset:*`int`, `?`*member:type2*`)`  
-   Access the field of type *type2* at address *struct* + *offset*.
+* `foreign lpvm access(`*struct:type*, *offset:*`int`, `*size:*`int`,
```
pschachte, line 10: You should be aware that the *struct* value may be tagged, so it may not point to the start of the struct, which means that *struct* might not point to the first byte of the structure.  This should be documented here, and it probably means that you'll also need to add a *start_offset* argument, like `mutate`,  and when you reuse it, the `alloc` instruction should return *struct* - *start_offset*.

If that's not actually a problem, the doc for `access` should mention that *struct* might be tagged, so it might not point to the first byte of the structure.

CaviarChen: Actually I'm a bit comfused about this part. 

My understanding is that the tag is combined with the pointer if it doesn't execeed a tag limit. Otherwise, the struct will contain a part to record the tag. So a mask is required to get the actual pointer? How come this won't break the bdw gc?
Why don't just make pointers pointing to the first byte of the structure and consider the tag as a field if there is one?
Is it true that an unboxed constructor can't exceed the tag limit?
Does the size contain the tag part?

I'm happy to discuss this in the next meeting if you think it's a bit complicated. This doesn't block any of my work.

pschachte:
Hi Zed,

> My understanding is that the tag is combined with the pointer if it 
> doesn't execeed a tag limit. Otherwise, the struct will contain a part 
> to record the tag.

That's right.

> So a mask is required to get the actual pointer?

That would work, but I'm just using arithmetic, because many CPUs can 
fold that arithmetic into the fetch instruction, so it has 0 cost.

> How 
> come this won't break the bdw gc?

Luckily, it was designed not to be sensitive to the low order bits.  I 
guess they were aware that some people use the low order bits as tags. 
Mercury does the same thing, and they use BDW, too.

> Why don't just make pointers pointing to the first byte of the structure 
> and consider the tag as a field if there is one?

That's less space and time efficient.

> Is it true that an unboxed constructor can't exceed the tag limit?

The number of tag bits is determined by the alignment of allocations. 
We've set that to 3 for 64 bit machines.  If there are more non-constant 
constructors than can be represented with that many bits, we set aside 
one tag (the highest one) to indicate that there is a secondary tag, and 
then the first word of the allocated structure stores the secondary tag.

> Does the size contain the tag part?

Yes, the size is the number of bytes for each constructor, including the 
secondary tag for structures that have them.  Note that different 
constructors for a type may have different sizes, so it's not enough to 
know the type of the structure, you need to know the constructor to know 
the size.  But of course, by the time you're accessing fields, you'll 
know the size.

> I'm happy to discuss this in the next meeting if you think it's a bit 
> complicated. This doesn't block any of my work.

Sure, we can discuss it Thursday if I've left things out.


-- 
Peter Schachte
Assistant Director, Infrastructure - IT
Melbourne School of Engineering
Senior Lecturer,
School of Computing and Information Systems
University of Melbourne
http://bit.ly/peterschachte

We promote flexible working at the Melbourne School of Engineering. I am 
currently working hours remotely. If you wish to contact me, my most 
preferred contact method is email.

I’m sending this message now because it suits me, but I don’t expect 
that you will read, respond or action it outside of your working hours.

CaviarChen:
Hi Peter,

I think the current code doesn't handle secondary tag correctly.
In this part of code:
https://github.com/pschachte/wybe/blob/507ad919d95c4a2b3f4c8a235faced47bdd47a98/src/Blocks.hs#L727-L747
It gets the baseAddr by subtracting the tagArg (start_offset).

However, in `Normalise.hs`, e.g. :
https://github.com/pschachte/wybe/blob/507ad919d95c4a2b3f4c8a235faced47bdd47a98/src/Normalise.hs#L686-L693
https://github.com/pschachte/wybe/blob/507ad919d95c4a2b3f4c8a235faced47bdd47a98/src/Normalise.hs#L635-L640
https://github.com/pschachte/wybe/blob/507ad919d95c4a2b3f4c8a235faced47bdd47a98/src/Normalise.hs#L599-L605

The start_offset is set to be the actual tag (even if it exceeds the tagLimit). The offset is also computed by using the actual tag. One of the access doesn't even consider the tag.

I think the start_offset should be `min(tag, tagLimit+1)` and the start_offset is the one used for computing the baseAddr. We should use `tag` instead of `start_offset` in access and mutate instruction? The tagLimt is a constant and using the `tag` can provide more information.

Also I am feeling that getting rid of `start_offset` or `tag` might not be that bad. We need the baseAddr only for non-destructive udpate and reusing dead cell, and one more `llvm and` instruction for those two cases doesn't seem to be that bad. But yes I think the `start_offset` or `tag` can let us have better code, It just that we need to be a lot careful with it.

Issue #96, opened 2020-10-12
Incorrect int comparison


CaviarChen:
```
?res = 10 > 5
!println(res)
?res = 10 < 5
!println(res)
```
gives output:
```
false
false
```

and for this program
```
!read(?a:int)
!read(?b:int)
if { a > b ::
    !println("a > b")
| otherwise ::
    !println("a <= b")
}
```
we have:
```
> ./test 
5 10
a <= b
> ./test 
10 5
a > b
> ./test 
5 -10
a <= b
> ./test 
-10 5
a > b
```

CaviarChen:
I think https://github.com/pschachte/wybe/commit/6536899d1e0790e2c4a25592f0ba5df4cf11ce62 only fixed the first problem but not the second one?

pschachte:
That's a completely different bug.  Could you please open a new one for that?  I've just committed a new test case test-cases/final-dump/read_int.wybe that shows that this bug only occurs when the numbers are read from stdin.  Using the same numbers a program constants, with inlining turned off, it works fine.  Take a look at the generated LLVM code.  I'm going cross-eyed trying to see what's wrong with it.  I marked the generated code correct, because as far as I can see, it is.

PR #97, opened 2020-10-12
Use `int64_t` instead of `int` in Wybe's C runtime.

CaviarChen:
Wybe's int is 64-bit (since we only support 64-bit system right now), but `int` is 32-bit (in most cases), so it caused a mysterious bug with io of negative numbers and signed comparison (the second problem mentioned in #96 ).

I choose the fixed width integer types becuare wybe's int can only be 64-bit right now, and I also found the normal integer types of C are annoying as their size may vary even on system with the same bit width. As far as I known, `long` could be 64-bit or 32-bit on 64bit machine with different C compilers.


pschachte (approved):
Nice catch.  It never occurred to me it was a bitwidth problem.

Issue #98, opened 2020-10-12
bitwidth warning compiling cbits.c


pschachte:
When compiling cbits.c, I get these warning messages:
```
wybelibs/cbits.c:16:35: warning: format specifies type 'long' but the argument
      has type 'int64_t' (aka 'long long') [-Wformat]
    return (int64_t)printf("%ld", x);
                            ~~~   ^
                            %lld
wybelibs/cbits.c:28:42: warning: format specifies type 'long' but the argument
      has type 'int64_t' (aka 'long long') [-Wformat]
  return (int64_t)fprintf(stderr, "%ld", x);
                                   ~~~   ^
                                   %lld
wybelibs/cbits.c:51:18: warning: format specifies type 'long *' but the argument
      has type 'int64_t *' (aka 'long long *') [-Wformat]
    scanf("%ld", &x);
           ~~~   ^~
           %lld
```

CaviarChen:
This is awkward, because in the clang on Ubuntu, `int64_t` is `long`.
```
clang  -I /usr/local/include -c "wybelibs/cbits.c" -o "wybelibs/cbits.o"
wybelibs/cbits.c:16:36: warning: format specifies type 'long long' but the argument has type 'int64_t' (aka 'long') [-Wformat]
    return (int64_t)printf("%lld", x);
                            ~~~~   ^
                            %ld
wybelibs/cbits.c:28:43: warning: format specifies type 'long long' but the argument has type 'int64_t' (aka 'long') [-Wformat]
  return (int64_t)fprintf(stderr, "%lld", x);
                                   ~~~~   ^
                                   %ld
wybelibs/cbits.c:51:19: warning: format specifies type 'long long *' but the argument has type 'int64_t *' (aka 'long *') [-Wformat]
    scanf("%lld", &x);
           ~~~~   ^~
           %ld
```

C is so weird. I'll try to figure out what is the best practice.

jimbxb:
@pschachte can this issue be closed?

pschachte:
Yup.

PR #99, opened 2020-10-12
Fix bitwidth warning compiling cbits.c. #98

CaviarChen:
#98 
According to: https://en.cppreference.com/w/c/types/integer


pschachte (approved):
I don't think this is the right fix, but I'll commit it anyway and open a new issue.

PR #102, opened 2020-11-26
Avoid overlapping-patterns warning && update test cases.

### LINKED ISSUE #80: "adjustNth refers beyond list end" raised in Callers.hs
int_list.wybe
```
pub type int_list { pub [] | [|](head:int, tail:int_list) }


pub def print(x:int_list) use !io {
    if { x = [ ?h | ?t] :: 
        !print(h)
        !print(" ") # the problem seems related to this one.
        !print(t)
    }
}

pub def println(x:int_list) use !io {
    !print(x)
    !nl
}

?x = [1,2,3]
!println(x)
```

```
> wybemk int_list
adjustNth refers beyond list end
  error, called at src/Callers.hs:79:20 in main:Callers
```

CaviarChen:
This PR includes:

1.  Add Placeholder item for `CallProperty`, `CallSiteProperty ` and `InterestingCallProperty`.
Otherwise there will be overlapping-patterns warning for cases like https://github.com/pschachte/wybe/blob/19556330d5cf7f301794750ba22f1a93b9243930/src/Transform.hs#L51-L55
Not sure if this approach is a good one. It changes the hash of specializations.

2. Update a test case since we fixed #80.

3. Remove a test case since it is already covered in complex case: testcase_multi_specz-int_list.


pschachte (changes requested):
Looks fine, but please put in a little more doc for each of these placeholders explaining why it's needed (or explain it in one place and reference that at each placeholder).

pschachte (approved):

Issue #122, opened 2021-03-23
VS Code integration utterly fails.


pschachte:
Attempted support for colourisation is in `plugins/vscode` directory.

CaviarChen:
Wow! There are so many exciting things going on with Wybe.
Sorry that I don't have time to contribute (at least recently), but I'm here to share some ideas. I actually tried to build a vscode plugin for Wybe before and failed.

In short, I think we could try to build a language server directly.
Here are some reasons:
1) rewrite the parser in Textmate is annoying and I don't think the wybe syntax is stable now (correct me if wrong) so we need to maintain two parsers.
2) Textmate syntax is not powerful enough and we need to switch to language server sooner or later.
3) We can build the language server in Haskell so we can reuse the existing parser. Haskell has a pretty good lsp implementation: https://github.com/haskell/lsp
For highlighting, all we need is a function that takes a wybe source file and returns a list (or tree) of symbols with their range in the text and types (supporting Document Symbols Request: https://microsoft.github.io/language-server-protocol/specifications/specification-3-15/#textDocument_documentSymbol). 

Potential problems:
1) There might be a performance issue since our native implementation will analyze the whole file every time we change it.
2) Might be hard to support recovering from syntax errors and producing a resilient parse tree.

pschachte:
Great to hear from you, Zed.

I was really just trying to get colourisation working.  I was hoping I could put in an hour or two to get at least comments coloured, and maybe proc names and even variable assignments.  I spent 4 or 5 hours and nothing worked beyond recognising that a filename ending in .wybe was a Wybe file.  But somehow when you type the open parenthesis after a proc name, it shifts the whole line to the left margin.

I suppose a language server is a good idea in the long run, as it has much more potential, such as highlighting type or other semantic errors.  And you're right, the syntax is not stable yet, and being able to use the same parser for the language server and the compiler would be good design.  On the other hand, it has the problem that a syntax error in the middle of the file will mess up everything that comes later.  But the biggest problem is that it looks like a lot of work, and it's just not a very high enough priority.

CaviarChen:
I think the one I made works with comments and can highlight keywords (exact match).
I'll try to find it and create a PR later (hopefully today, or at least by the end of this week).

pschachte:
That would be great, thank you!

CaviarChen:
I closed the previous pull request.
I implemented a version and put it here: https://github.com/CaviarChen/vscode-wybe
[...]

PR #123, opened 2021-03-24
Add a naive vscode plugin implementation

CaviarChen:
Related issue #122.

I think the existing version is using a legacy template. Mine is based on this guide: https://code.visualstudio.com/api/language-extensions/syntax-highlight-guide

I tried some of the rules in the existing one but they don't work, so I just replaced them.


pschachte (changes requested):
This looks really good, but can you please update the README.md file to add installation instructions?

CaviarChen:
Sorry for getting back to this so late.
The guide can be found here: https://code.visualstudio.com/api/working-with-extensions/publishing-extension
If you want to publish this extension, maybe it is unnecessary to describe this in README?

pschachte:
Oh, wow!   I had no idea how complicated it would be!   Yes, I definitely think there should be really simple instructions, hopefully along the lines of "copy this directory to such-and such a place and restart VS Code" or "run this command in the VS Code command palette".  I hope you don't have to install JavaScript and node.js just to get code colouring for Wybe code.  What did you have to do to test your extension?

CaviarChen:
**For testing:**
Open folder `plugins/vscode` in vscode and press F5. A new vscode window will pop up with this extension enabled.

**For publishing it:**
I don't think it's a good idea to let users build the extension themselves, it requires node.js etc.
We can build and pack the extension into a `*.vsix` file, but it is still not that easy to use and we need to figure out a way to distribute the `.vsix` file.

I think the best approach is to publish this extension to vscode marketplace properly. The main job here is to create an account following that guide. If you don't mind, I can do that for you, but just that I'll have the control of it, or you can create it and give me a token or something, basically give me the permission to manage it in marketplace.
Once have the account, it is easy to build and publish (upgarde) the extension to vscode marketplace. I can run the command manually for now, just let me know when someone update the code for the extension.

I can also setup a github action to do the last step automatically, but since we won't change it frequently at least for now, that's something for the future.

PR #126, opened 2021-04-14
Remove unused return type of functions in Blocks.hs

09milk:
Return type of following functions had been changed to Unit:
```
cgen
cgenLLVMBinop
cgenLLVMUnop
cgenLPVM
addInstruction
```


pschachte (changes requested):
`assign` already returns `Codegen ()`, so for all the function that end with a call to `assign` and then do `return ()`, you should just drop the `return ()`.  Also `zipWith3M_` in this context returns `Codegen ()`, so you can drop the `return ()` at the end of `addInstruction`.  You may find other places you could get rid of `return ()` because the previous thing already returns `Codegen ()`.  Please fix these things and resubmit the PR.

pschachte (approved):

PR #129, opened 2021-06-10
Fix make install command.

CaviarChen:
The current `make install` is broken on both ubuntu and macos (zsh).
This RP fixes that and makes the CI to test `make install`.


pschachte (approved):
Thanks for that!

Issue #49, opened 2020-05-06
Inconsistent code logic between LLVM IR and executable


09milk:
Using [bigloop.wybe](https://github.com/pschachte/wybe/blob/master/test-cases/execution/bigloop.wybe) as an example, output executable using command:
`wybemk bigloop`
will get an executable doing **exactly** what the code say: looping over `i` and add 2 to `sum` each time.
`wybemk bigloop.ll`
will get a LLVM IR output with fully optimized code without any looping.

As a side note, when disassembling the bigloop executable I also noticed the fact that the final executable will include all standard types and functions, including those it never used during runtime.
```
$ wybemk --version
wybemk 0.1 (git 40f3b16)
```

PR #131, opened 2021-07-02
Fix inconsistency between .ll and executable 

09milk:
As described in #49, this pull request fix the issue by enabling LLVM optimisation pass manager on executable.

A similar commit 1d45ba686a had been made by @pschachte to enable the pass manager on object file, but inconsistency between `makeWrappedBCFile` and `makeWrappedObjFile`  was introduced by 1d45ba686a on 17 Oct 2020. #49 was reported on 6 May 2020, well before the commit. 

What we are doing seems like a band-aid to the problem instead of fixing the issue behind it, but at least this pull request will make the compiler use LLVM optimized code to generate the executable instead of unoptimized code.

### DIFF
--- a/src/Emit.hs
+++ b/src/Emit.hs
@@ -169,35 +169,14 @@ withModule llmod action =
 -- Target Emitters                                                        --
 ----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


pschachte (approved):
Thank you, that's very helpful.

Issue #134, opened 2021-07-17
Bug in multiple specialisation


pschachte:
Compiling this code in the latest master branch gives a multiple specialisation error:
```
pub constructor lower:int .. upper:int

pub def {test} [?head:int | ?tail:_]:_ = ?seq where {
    seq^lower <= seq^upper
    ?head = seq^lower
    ?tail = (seq^lower+1)..seq^upper
}

for ?i in 1..10 {
    !println(i)
}
```
(Note that the type `_` is an alias for the type defined in the current module.)
Compiling this gets the error message:
```
EncodeException "reference to undefined global: Name \"int_sequence.[|]<0>[785a827a1b]\""
```
When compiled with `--no-multi-specz`, it compiles fine and works as expected.

CaviarChen:
This bug is a bit tricky and it took me quite long time to figure out. (Or maybe I just forgot a lot about Haskell and Wybe).
I believe there are two bugs.

1.
https://github.com/pschachte/wybe/blob/ae35692a8e2c77ba32beb1ffb8693c708f16d798/src/Transform.hs#L241-L242
The `orderedProcsTopDown` is wrong. 
In your given case (`bug` is the module name), its value is
` [AcyclicSCC bug.<0>,AcyclicSCC bug...<0>,AcyclicSCC bug...<1>,AcyclicSCC bug.=<0>,AcyclicSCC bug.gen$1<0>,AcyclicSCC bug.[|]<0>,AcyclicSCC bug.lower<0>,AcyclicSCC bug.lower<1>,AcyclicSCC bug.upper<0>,AcyclicSCC bug.upper<1>,AcyclicSCC bug.~=<0>]`
Note that `bug.gen$1<0>` should be self recursive but it is marked as `AcyclicSCC` here. That mislead the multiple specialization expansion so one of the required version is skipped.
I haven't got time to dig into `getSccProcs`, @pschachte can you have a look?

2.
https://github.com/pschachte/wybe/blob/ae35692a8e2c77ba32beb1ffb8693c708f16d798/src/Transform.hs#L273-L280
This should be `Set.difference required' required`, the current one is always empty. I'll create a PR to fix this tomorrow.

CaviarChen:
Figured out what's going on with bug 1. Writing the fix now.

pschachte:
Zed's commit #137 fixes this.

PR #137, opened 2021-07-18
Fix multiple specialization bug #134

CaviarChen:
Detail can be found in #134


pschachte (approved):
Looks good.  Thanks, Zed!

Issue #130, opened 2021-06-10
Build failure caused by multiple specialization on installed wybelib


CaviarChen:
By default, the wybelib is intalled in a folder that the normal users do not have write permission.
If the compiler tries to generate a specialized version for a procedure in stdlib, it fails.
Now, our stdlib is complex enough to have this problem.
Example:
```
$ cd test-cases/execution
$ wybemk echo
FdStreamException "Permission denied"
```

I think in this case, the compiler should copy the object file to a local folder, such as `./_build/`.

We should also consider to cover this in our CI by using the installed wybemk to run all tests.

pschachte:
This is a really tricky problem.  It doesn't sound ideal to have the compiler make local copies of whole library modules.  As the library grows, and particularly when only a few procs in many large library modules need specialisation, this could get expensive.  What about generating a single local file of specialised versions of any library (or generally non-overwritable object) code?  This would be an exception to the one-one relationship between object files and source files.  But I actually already broke that a little in fixing support for module directories (I generate a separate object file for whatever code and metaadata is contained in the directory module).

CaviarChen:
I personally don't think this is that tricky.
It's all about granularity, we can make a copy per project (like I mentioned) or per user (put it in somewhere like `~/.wybe/cache`). We can have a separate file for specialized versions or even for each procedure.
Switching between approches is also not that difficult. I recommend we just take whatever that is reasonable and easy to implement.
Mainly I am kind of against optimizing (and spend time on it) too early, we can always improve it when it become a real problem.
Based on sizes current wybe programs, performance of cpu (also we do support incremental build) and cost of disk space, the compiler architecture could be a lot different when that happens.

pschachte:
Fixed in commit #139

PR #139, opened 2021-07-31
Introduce local cache file which allows multiple specialization on read-only libs

CaviarChen:
By the default install config, wybemk doesn't have write permissions on object files of the stdlib. However, the multiple specialization framework needs to update those file. So now we have a problem #130.

This PR introduces "local cache file": if wybemk tries to write to a file that it doesn't have write permission, it will create a cache file under `~/.wybe/cache` and uses it instead. It also records the hash of the original file so the cache file will become invalid once the origincal file has changed. The whole thing is transparent to most parts of the compiler.

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -232,7 +232,7 @@ import           Transform                 (transformProc,
 import           Types                     (typeCheckModSCC,
                                             validateModExportTypes)
 import           Unbranch                  (unbranchProc)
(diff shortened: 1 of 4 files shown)


pschachte (approved):
Nice solution to the problem.  Thanks for that!

Could you add some documentation of what the `.meta` file is for?

PR #143, opened 2021-08-10
fix for single-line comment at eof not parsing

jimbxb:
Example program that fails to parse:
```
!println("Wominjeka!")
#oops!
```
(with no terminal new line)

Output before fix:
```sh
% wybemk test
test.wybe:2:1: Syntax error: unexpected Unterminated comment begins here
expecting simple expression
```


pschachte (approved):
Looks good, thanks for the fix.

Issue #145, opened 2021-08-16
`foreign lpvm alloc` only supports `int` constant size


jimbxb:
The following code should compile, however does not.

```
def {noinline, semipure} foo(size:int) {
    foreign lpvm {impure} alloc(size, ?mem:machine_word)
}
```

Instead, the following error message is produced:
```
src/AliasAnalysis.hs:654:9-42: Non-exhaustive patterns in [ArgInt size _, struct]
```

CaviarChen:
Right, that part of code is related to the dead memory cell reusing. At the time we implemented that, we only allowing calling alloc with constant size. 

A easy fix could be just ignoring the variable size case. 

I think there was a discussion about supporting a subset of the variable size case for immutable array, but I don't know how that went. In general, the dead memory cell reusing doesn't need to know the actual size of a cell in compile time, all it needs is whether two cells have the same size.

jimbxb:
I am going to close this issue as I believe is now resolved.

PR #146, opened 2021-08-16
Fix for alloc args (#145)

jimbxb:
This fix changes code related to dead cell reuse (cc @CaviarChen )


pschachte (approved):
Great, thanks for fixing that!

Issue #149, opened 2021-08-27
Type checking fails when multiple calls are made to a generic type in the same context with different concrete types.


jimbxb:
Example:
```
def foo(x:?T) use !io {
    !println("in foo")
}

!foo(1)
!foo('a')
```

Error:
```
Error detected during type checking of module(s) test
test.wybe:6:2: Type error in call to foo, argument 1
```

jimbxb:
On a similar vein, LLVM serialisation issues occur if this is fixed. When calls are made to a generic function with differing sizes this error occurs:

```
def foo(x:?T) use !io {
    ?l = [x]
    !println(length(l))
    !println("nice")
}

!foo(1)
!foo('a') 
```
```
EncodeException "The serialized GlobalReference Name \"test.foo<0>\" has type PointerType {pointerReferent = FunctionType {resultType = VoidType, argumentTypes = [IntegerType {typeBits = 64}], isVarArg = False}, pointerAddrSpace = AddrSpace 0} but should have type PointerType {pointerReferent = FunctionType {resultType = VoidType, argumentTypes = [IntegerType {typeBits = 8}], isVarArg = False}, pointerAddrSpace = AddrSpace 0}"
```

pschachte:
I've looked into this enough to know that your diagnosis is right:  the binding for type variable `T` is stored in the typing, which it shouldn't be.  But not enough to fix it.

pschachte:
Fix committed.

PR #150, opened 2021-08-29
Refreshed type variables when type checking calls (#149)

jimbxb:


pschachte (approved):
Looks good, thanks for that!

Issue #153, opened 2021-09-08
Allow constructor declarations to omit field names for some or all fields


pschachte:
Then don't generate accessors and mutators for anonymous fields.

pschachte:
Closed by #154.

PR #154, opened 2021-09-10
Allowed constructors to omit field names (#153)

jimbxb:
Also removed normalization to create accessors/mutators for said fields

### DIFF
--- a/src/Normalise.hs
+++ b/src/Normalise.hs
@@ -12,7 +12,7 @@
 module Normalise (normalise, normaliseItem, completeNormalisation) where
 
 import AST
-import Config (wordSize, wordSizeBytes, availableTagBits, anonFieldName,
+import Config (wordSize, wordSizeBytes, availableTagBits,
                tagMask, smallestAllocatedAddress)
 import Control.Monad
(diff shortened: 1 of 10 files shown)


src/Normalise.hs
```diff
@@ -634,7 +656,7 @@ constructorItems ctorName typeSpec params fields size tag tagLimit pos =
         pos]
 
 
--- |Generate deconstructor code for a non-const constructor
+-- |Generate deconstructor code for a non-const deconstructor
```
pschachte, line 659: 'constructor' was the right word here; please change back.

pschachte (changes requested):
I think it would be nicer to leave out the _: in this case.  Ie, a constructor argument without a : would be interpreted as specifying just the type of the argument.  I know that's different from the way that is interpreted when it appears in a function prototype (which is basically what this is), but the contexts are different (a constructor with no type for an argument makes no sense), so I don't think this would be confusing.

Also, please update the language doc to document this feature.

pschachte (approved):
Looks good, thanks.

PR #162, opened 2021-09-23
Added a more usable string type

jimbxb:
Not yet implemented: non-C-style special resources


wybelibs/wybe/control.wybe
```diff
@@ -11,13 +11,13 @@ pub def {terminal,semipure} exit(code:int) {
 
 # Print the source location and and error message, and abort the computation.
 
-pub def {terminal,semipure} error(message:string) use call_source_location {
+pub def {terminal,semipure} error(message:c_string) use call_source_location {
```
pschachte, line 14: There should be a Wybe string version of error.

wybelibs/wybe/string.wybe
```diff
@@ -1,27 +1,145 @@
...
-use wybe.char
+use wybe.int, wybe.c_string, wybe.range
+use wybe.int, wybe.char, wybe.bool
+use wybe.io, wybe.float, wybe.count
+use wybe.comparison
 
-representation is address
+constructor empty
+          | buffer(len:int, raw:c_string) # must be first non-const ctor
+          | concat(left:_, r:_)
```
pschachte, line 10: Please rename `r` to `right`

pschachte (changes requested):
Looks really good, just a few small changes needed.

wybelibs/wybe/control.wybe
jimbxb: I agree with this. 

However, I have tried to implement this via the following
```
pub def {terminal,semipure} error(message:string) use call_source_location {
    foreign c {terminal,semipure} error_exit(call_source_location, c_string(message))
}
```
And received this error message: 
```
wybelibs/wybe/control.wybe:14:68: error has normal (total) determinism, but declared terminal
```

It appears there is some underlying issue with determinism checks of terminal procedures. Yet to try to diagnose further

pschachte (approved):

pschachte:
Uggg!  There are conflicts in four .exp files because of the change I just committed for resources.  Could you please re-run ./update-exp and commit the changes to this PR, so I can merge it?

jimbxb:
On it now. 

Looks like string now needs to import `wybe.phantom`, which wasn't the case before when importing `io`

pschachte (approved):

Issue #166, opened 2021-09-30
Change the notation for type variables


jimbxb:
Currently the notation for type variables makes use of a preceding `?`, which of course is used to denote outward flow also. 

In some situations (such as a higher order types declaration), where types and flows must be handled in tandem, the notation become confusing: is `?a` an outflowing value of type `a` or is an inflowing value of type `?a`.

On solution to this is to use a `:` separator to disambiguate these cases.

Alternatively, we can change the prefix symbol for type variables. One possibility is to follow other languages that use a `'`, but this may be confused in the scanner for a character quote. Else we could use a `^` as this would be unused (yet reserved) in this context. I'm in favour of the `^`

pschachte:
How about defining a type variable as a single upper case letter followed by 0 or more digits?  The forbid type names of that form.  That has the advantage of following common conventions for type variables, and not requiring any special prefix symbol.  

Another option would be to explicitly "quantify" type variables.  That already happens in type declarations, where type variables appearing in the type definiton must appear as parameters of the type being defined.  We could provide a syntax for quantifying and scoping a type variable in other contexts, such as proc declarations.

jimbxb:
After further discussion the resolution of this will be to use `[A-Z]\d*` (single uppercase followed by digits) as previously suggested.

PR #189, opened 2021-11-05
Changed the notation for type variables

jimbxb:
This PR concerns issue #166 

Type variables are now of the form `[A-Z][0-9]*`, and types/modules cannot be declared with these names.

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -1039,10 +1039,10 @@ be placed in a Wybe source file to define a linked list type with whatever name
 is deemed suitable.
 
 ```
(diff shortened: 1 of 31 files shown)


pschachte (approved):
All good.  Thanks for doing that.

There's one tricky case that I didn't see handled or documented:  what about a file with a name like `X.wybe` or `X123.wybe`?  That would create a module with an invalid name.  Could you please also document that that's not a valid Wybe source file name, add a test to report an error if someone compiles such a file (including if it's in a directory being compiled), and add a test case `final_dump/X.wybe`?  Thanks.

PR #201, opened 2021-11-09
Parser cleanup

### LINKED ISSUE #200: Sub-module type constuctors do not parse correctly
Example:
```
type foo { pub bar(int) }
```

Compiler errors, saying that parameters are untyped.

jimbxb:
`StmtExpr` always felt awkward to me. I've changed it (and appropriate other cases) to use `Term` in place, which I think captures the concept in a less awkward name. 

Also fixes #200 with an appropriately updated test case.


pschachte (changes requested):
Sorry, I accidentally committed a change to master adding a case statement, when I meant to commit it to my branch.  I tried to reconcile this PR with that commit, but failed, and this PR no longer compiles.  Can you please fix it?  I think the necessary fix should be clear (stmtExpr -> term).

jimbxb:
No worries. Working on master is always risky business. I'll fix it up :)

Cool stuff with the `case` too. Doesn't affect much further down the line which makes my life a little easier

PR #211, opened 2021-11-26
Refactored Resources

### LINKED ISSUE #209: `use` blocks create variables that are "uninitialized" by LPVM generation time
Use blocks create temp variables that are used to (re)store the current value of an initialised resource. These temp variabless are not present in the `VarDict`s of `Cond` stmts (etc), and as such cause issues when translating to LPVM. The solution should be to add these variables and their types to the enclosed `VarDict`s.

Example:
```
resource counter:int = 0

use counter in {
    if { false :: pass }
    !nl
    !nl
}
```

Error:
```
Error detected during generating low level code in resource_unbranch
Uninitialised variable 'tmp#0'
```

jimbxb:
This fixes #209 

- the process of transforming resources into arguments is now handled with a `StateT`, with the state storing information about tmp variables 
- additional test case for #209

### DIFF
--- a/src/Resources.hs
+++ b/src/Resources.hs
@@ -33,7 +33,7 @@ resourceCheckMod _ thisMod = do
     reenterModule thisMod
     resources <- getModuleImplementationField (Map.toAscList . modResources)
     (chg,errs,resources') <-
-        fmap unzip3 $ mapM (uncurry checkResourceDef) resources
+        unzip3 <$> mapM (uncurry checkResourceDef) resources
     updateModImplementation (\imp -> imp { modResources =
                                               Map.fromAscList resources'})
     reexitModule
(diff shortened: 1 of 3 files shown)


src/Resources.hs
```diff
@@ -89,11 +89,27 @@ canonicaliseResourceFlow :: OptPos -> ResourceFlowSpec
...
-    return $ spec { resourceFlowRes = fst resTy}
+    return $ spec { resourceFlowRes = fst resTy }
 
 
 --------- Transform resources into variables and parameters ---------
 
+-- | Data type to store the necessary data for adding resources to a proc
+data ResourceState = ResourceState {
+    resTmpCtr    :: Int,
+    resTmpVars   :: Map VarName TypeSpec
```
pschachte, line 100: The type `VarDict` in AST is defined as `Map VarName TypeSpec`, so please use that alias here.

pschachte (changes requested):
All good except one small change.

pschachte (approved):
Thanks.

PR #235, opened 2021-12-14
Attempt at checking constraints

jimbxb:
This is an attempt to ensure that constant expressions with type constraints have valid type constraints. 

Notes: 
* Type constraints are ignored for constants in foreign code. 
  * `foreign llvm move` instructions are one exception to this rule.
* Strings can only be constrained to their respective types, `string` and `c_string`

~~The `BodyBuilder` part with simplifying a cast instruction is admittedly quite a hack. The reason for its existence is the large number of casts I had to introduce for ADTs. If there's a better way to do this or the casts in Normalise then I am happy to change this.~~

~~One alternative approach to this is to ignore type constraints to constants in foreign code, as the only place I've had to introduce these casts is in foreign code. Foreign code is already the wild west of types (looking at you, C), so leaving this case out shouldn't hurt, yet we still get to check constraints in non-foreign code.~~

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -424,7 +424,7 @@ typeErrorMessage (ReasonBadConstraint caller callee argNum exp pos) =
         "Type constraint (:) in call from "
         ++ prettyName "top-level" id caller
         ++ " to " ++ callee ++ ", argument " ++ show argNum
-        ++ ", is incompatible with expression " ++ show exp ++"."
+        ++ ", is incompatible with expression " ++ show exp
 typeErrorMessage ReasonShouldnt =
     Message Error Nothing "Mysterious typing error"
 typeErrorMessage (ReasonActuallyPure name impurity pos) =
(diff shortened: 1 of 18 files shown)


src/Types.hs
```diff
@@ -343,95 +345,103 @@ typeErrorMessage (ReasonAmbig procName pos varAmbigs) =
...
 --     Message Warning pos $
---     "Proc " ++ name ++ " should be declared terminal"
+--         "Proc " ++ name ++ " should be declared terminal"
 -- typeErrorMessage (ReasonUnnreachable name pos) =
 --     Message Warning pos $
---     "In proc " ++ name ++ ", this statement is unreachable"
+--         "In proc " ++ name ++ ", this statement is unreachable"
+
 
+prettyName :: String -> (ProcName -> String) -> ProcName -> String
```
pschachte, line 442: I think AST.showProcName does what prettyName does; if possible, it would be better to use that.

pschachte (changes requested):
Also, could you please add a test case for the error messages?

jimbxb:
I want to improve the error message. Please hold off on merging :)

pschachte (approved):

Issue #223, opened 2021-12-09
Unique types should not be allowed as input/outputs in tests


pschachte:
When a test fails, all variables return to their previous state.  Since currently unique types can't be passed as inputs unless they're passed back out, that means unique inputs in a test call is not actually unique.

Additionally, it should not be possible to call a det proc with a unique in/out parameter in semidet context, ie, in the body of a semidet proc, or in a condition of a conditional, or in a disjunction, conjunction, or negation.

pschachte:
Fixed in #236.

PR #236, opened 2021-12-17
Uniqueness improvements

pschachte:
Make test (and failure) procs with unique arguments/resources illegal.

Make assignment to unique variables/resources in a test or failure context illegal.

Closes issue #223.

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -1435,8 +1435,9 @@ data ModuleInterface = ModuleInterface {
                                      -- ^The procs this module exports
     pubSubmods   :: Map Ident ModSpec, -- ^The submodules this module exports
     dependencies :: Set ModSpec,      -- ^The other modules that must be linked
-    typeModifiers :: TypeModifiers   -- ^The extra information of the type
-    }                               --  in by modules that depend on this one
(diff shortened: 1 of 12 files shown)


src/AST.hs
```diff
@@ -1435,8 +1439,9 @@ data ModuleInterface = ModuleInterface {
                                      -- ^The procs this module exports
     pubSubmods   :: Map Ident ModSpec, -- ^The submodules this module exports
     dependencies :: Set ModSpec,      -- ^The other modules that must be linked
-    typeModifiers :: TypeModifiers   -- ^The extra information of the type
-    }                               --  in by modules that depend on this one
+                                      --  in by modules that depend on this one
```
jimbxb, line 1442: This line should be swapped with the line below

src/Unique.hs
```diff
@@ -20,28 +20,40 @@ import Data.Set as Set
 import Data.Map as Map
 import Data.Maybe
 import GHC.TypeNats (Mod)
+import System.Posix.Internals (fdType)
```
jimbxb, line 23: Is this import intentional?

PR #244, opened 2021-12-29
fix more cases of bug 228

### LINKED ISSUE #228: Available resources are not handled correctly.
Resource availability is not handled correctly throughout the mode checking process.

Example:
```
def foo(?s:string) {
    ?io = phantom
    !read(?s)
}

!println(foo)
```

This program should report that the resource `io` is not available in the call to `read`. However, this compiles fine due to there being a variable named `io` defined, echoing the first line read from stdin.

pschachte:
canonicalise resource specs in use statements.  Check for type conststency of variables promoted to resources.

Closes #228.  I hope.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -144,7 +144,7 @@ data RoughProcSpec = RoughProc {
 } deriving (Eq,Ord)
 
 instance Show RoughProcSpec where
-    show (RoughProc mod name) = maybeModPrefix mod ++ name
+    show (RoughProc mod name) = maybeModPrefix mod ++ showProcName name
 
 
 -- |Type check a single module dependency SCC.  
@@ -226,6 +226,8 @@ data TypeError = ReasonParam ProcName Int OptPos
                    -- ^Output param not defined by proc body
                | ReasonResource ProcName Ident OptPos
                    -- ^Declared resource inconsistent
+               | ReasonUseType Ident OptPos
+                   -- ^Type of resource in use stmt inconsistent with other use
                | ReasonUndef ProcName ProcName OptPos
                    -- ^Call to unknown proc
                | ReasonArgType ProcName Int OptPos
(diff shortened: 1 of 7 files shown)


src/Types.hs
```diff
@@ -1764,15 +1770,21 @@ modecheckStmt m name defPos assigned detism tmpCount final
            ,postLoopBindingState assigned assigned',tmpCount')
 modecheckStmt m name defPos assigned detism tmpCount final
     stmt@(UseResources resources _ stmts) pos = do
+    resources' <- lift
+                  $ mapM ((fst <$>) . canonicaliseResourceSpec pos "use stmt")
+                    resources
+    -- Odd doing this bit of type checking during mode checking, but it would be
+    -- a bit messy to do this during type checking.
+    mapM_ (addResourceType (flip ReasonUseType pos)) resources'
```
jimbxb, line 1778: I think we should do this earlier. This *may* modify the ultimate type of some generic types, and those generic types could have been used previously

src/Types.hs
```diff
@@ -1764,15 +1770,21 @@ modecheckStmt m name defPos assigned detism tmpCount final
...
-            List.foldr Set.insert (bindingResources assigned) resources }
+            List.foldr Set.insert (bindingResources assigned) resources' }
     (stmts', assigned'', tmpCount')
         <- modecheckStmts m name defPos assigned' detism tmpCount final stmts
     let boundRes = USet.intersection (bindingVars assigned)
-                   $ USet.fromList $ resourceName <$> resources
+                   $ USet.fromList $ resourceName <$> resources'
     return
-        ([maybePlace (UseResources resources
+        ([maybePlace (UseResources resources'
```
jimbxb, line 1787: I think we need to modify the binding state as well here. If a use block binds a previously unbound variable corresponding to a resource, this variable should be unbound after the use block

jimbxb (changes requested):

jimbxb:
Closing this due to #243 and #246

PR #246, opened 2021-12-29
Final #228 changes

jimbxb:
These changes are mostly from @pschachte's #244 fix, and some light cosmetic changes.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -13,6 +13,7 @@ module Types (validateModExportTypes, typeCheckModSCC) where
 import           AST
 import           Control.Monad
 import           Control.Monad.State
+import           Control.Monad.Extra (ifM)
 import           Data.Graph
 import           Data.List           as List
 import           Data.Map            as Map
@@ -20,6 +21,7 @@ import           Data.Maybe          as Maybe
 import           Data.Set            as Set
 import           UnivSet             as USet
 import           Data.Tuple.Select
+import           Data.Function (on)
 import           Data.Foldable
 import           Data.Bifunctor
 import           Options             (LogSelection (Types))
@@ -28,7 +30,6 @@ import           Util
 import           Snippets
 import           Blocks              (llvmMapBinop, llvmMapUnop)
 import           Unique
-import Data.Function (on)
 import           Debug.Trace
 
 
@@ -315,7 +316,8 @@ typeErrorMessage (ReasonParam name num pos) =
         ", parameter " ++ show num
 typeErrorMessage (ReasonOutputUndef proc param pos) =
     Message Error pos $
-        "Output parameter " ++ param ++ " not defined by proc " ++ show proc
+        "Output parameter " ++ param ++ " not defined by proc " 
+        ++ showProcName proc
 typeErrorMessage (ReasonArgType name num pos) =
     Message Error pos $
         "Type error in call to " ++ name ++ ", argument " ++ show num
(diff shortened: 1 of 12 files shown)


src/Types.hs
```diff
@@ -307,17 +309,14 @@ instance Show TypeError where
 
 -- |Produce a Message to be stored from a TypeError.
 typeErrorMessage :: TypeError -> Message
-typeErrorMessage (ReasonParam name num pos) =
+typeErrorMessage (ReasonParam name num pos) =
```
pschachte, line 312: Gratuitous whitespace addition.

src/Types.hs
```diff
@@ -307,17 +309,14 @@ instance Show TypeError where
 
 -- |Produce a Message to be stored from a TypeError.
 typeErrorMessage :: TypeError -> Message
-typeErrorMessage (ReasonParam name num pos) =
+typeErrorMessage (ReasonParam name num pos) =    
     Message Error pos $
         "Type/flow error in definition of " ++ name ++
+        "Type/flow error in definition of " ++ showProcName name ++
```
pschachte, line 315: Duplicated line.

test-cases/final-dump/bug228-type.exp
```diff
@@ -0,0 +1,4 @@
+Error detected during type checking of module(s) bug228-type
+[91mfinal-dump/bug228-type.wybe:5:2: Foreign call 'move' with unknown type in argument 2
```
pschachte, line 2: This is a pretty unhelpful error message.  The statement doesn't look like a foreign move instruction.  The second argument is a string constant, so the type is known.  And there's actually nothing wrong at that statement; the problem is actually in the following use statement, which requires an incompatible type.  That's the disadvantage of applying the resource types from the `use` statements before processing the types from the calls:  the `use` statements always look like they've appeared first.

I started working on a fix for this that merges the types from `use` statements in with the calls in left-to-right order, but it becomes a pretty big change, and it's not clear it's worth it.

But it would be good if this error was at least reported as a call to `=` rather than `move`.

pschachte (changes requested):

src/Types.hs
jimbxb: We should add a pre-commit hook that strips whitespace

test-cases/final-dump/bug228-type.exp
jimbxb: I agree this is unhelpful. WE can remove this error by moving where we do the use type checking up a little, which I think is best for now. I'm not sure it is worth while either making such a large change for this, but perhaps we can keep that one in the books if it comes around again. Hopefully we don't get an issue 228 again...

As for the type of the second argument, when an `?out = in` is flattened, the out-flowing argument becomes argument 2 to the `move` instruction, so the argument that is reportedly untyped is `?bar` (with type `XXX`). We may consider flattening slightly differently, perhaps by adding a flag in the foreign `move` call that says this move instruction came from an assignment, then print an error accordingly.

pschachte: I think the most natural expectation of users is that a type error occurs in the first place where a variable is used with a different type than it previously had, counting the parameters as occurring first.  That means the best fix is to handle the `use` statements, together with all the calls, in top-to-bottom order.  Likewise the explicit casts and foreign instructions.  But I this would be a big change and the payoff will be small.  I'll put in an issue for it, so we don't forget, but I don't think it's a high priority.

You're right about argument 2 vs. argument 1.  But it's actually during type checking, specifically mode checking, where calls to `=` in in/out and out/in modes are turned into foreign move instructions.  This is to avoid having to generate those modes for the = proc for every single type.  It happens in the first equation for `modecheckStmt`, around line 1634-1639.  So we could improve the error message by adding another argument to `modecheckStmt` to indicate what the original statement was, and use that to construct the error message.

jimbxb: If we look in flatten, line 390, `flattenAssignment`, assignments get flattened into moves early on. I believe we need to do this because there's no `=` assignment defined for generically typed terms. We could special-case this, if we wanted, where an in-out or out-in `=` passes type checking iff both operands have valid types.

In it's current state, though, I believe we can remove the part of `modecheckStmt` that deals with assignment->move transformations as it is never reached due to the flattening that occurs, but I agree that if we ever reached that, then I agree with adding additional arguments to `modecheckStmt` to show how the original call appears.

jimbxb: I've made changes that remove the flattening of `=` into a move instruction, and instead special-cased assignment in type checking. This improves a fair few type errors where a call to `=` shows up as a move

jimbxb: A few more changes.

The change to delay the assignment->move transformation was to allow user-defined `=` to take precedence over the "built-in" assignment.

Error messages are also now no worse than before the previous commit. Of course, it still doesn't have the error at the use stmt, and instead at the `=` before, but I think this is okay for now.

pschachte (approved):
Nice solution!

PR #248, opened 2021-12-30
 Add support for hex character escapes.  Fixes #230.

### LINKED ISSUE #230: Allow for more escaped character constants
Examples that would be nice to support are the following:

* hex: `'\x...'`
* octal: `'\...'`
* ~~null: `'\0'` (this is a special case of octal)~~ (added in #231)

pschachte:


pschachte:
I haven't implemented octal character escapes.  I'm not sure if anybody uses them anymore, and they'll conflict with the null character escape (would "foo\012bar" be foo and bar separated by character 10, or foo and bar separated by a null character and "12"?).  I'll close #230 for now, but if you feel strongly about octal character escapes, and have a good solution to the conflict above, feel free to reopen it.  They wouldn't be hard to implement.

Closes #230.

jimbxb:
I think the way you could could solve this is by using a zero-padded octal escaped character. If you want null-12 you can use `\00012`; if you want 012_8 you can use `\012`; and if you happen to not use an octal character after you don't need to zero-pad it (such as `\0ab`). 

I believe we'd also have to have a parse error if one writes `\477` (generally above 255 = 377_8). We could also handle it silently, parsing the former as `\47` and `7`.

Either way, I also don't think we *need* octal. Hex is probably good enough

Issue #250, opened 2022-01-05
Failure calls can cause code to be incorrectly removed


jimbxb:
Given this example:

```
if { fail ::
    pass
| else :: 
    pass
}
!println(1)
```

We should expect that the `if` block can be removed, and the `!println(1)` to be executed, however, the program terminates without printing anything. The issue lies within mode checking and unbranching.

In mode checking a conditional, if the condition is guaranteed to succeed or fail, only the then or else branch is kept after the condition, so after type checking we get the following body:

```
fail
pass
!println(1)
```

When this body gets fed through the unbrancher, we meet a `fail`. A `fail` causes the unbrancher to produce no further statements, so we end unbranching with the following body:

```
fail
```

Which is why this program terminates without printing. 

To remedy this, we may consider changing how either `fail` (and more generally failure determinism) is unbranched or modify type checking to not mode check failure in that way. I think the better approach is to modify how `fail` is unbranched-- in a semidet context execute the `alt` stmts, in a failure context halt execution, else act like a nop. 

EDIT: This might not work though.

Consider: 

```
def {test} foo {
    if { fail :: pass | else :: pass }
}
```
would get mode checked into 
```
def {test} foo {
   fail
   pass
}
```

Which would always fail. We instead want this to always succeed. Perhaps modifying mode checking is the way to go, or a combination of the two

PR #258, opened 2022-02-25
Fix for bug #250

jimbxb:
This fixes issue #250 

In the case a condition is guaranteed to fail, we used to remove the cond, but this was erroneous in a `test` proc because the condition now causes failure of the whole proc. We also cannot remove the condition because it may have side-effects. Instead, we still generate the cond, but have nothing in the "then" branch. 

A relevant test case exists for this.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -2048,9 +2048,8 @@ modecheckStmt m name defPos assigned detism tmpCount final
                     $ bindingVars assigned3
         impurity <- lift $ stmtsImpurity tstStmt'
         let stmts' = if impurity > Pure 
-                     -- if the test is non-pure, need to keep the cond around
-                     then [maybePlace (Cond (seqToStmt tstStmt') [] elsStmts'
-                             (Just condBindings) (Just bindings) res) pos]
+                     -- if the test is non-pure, need to keep the test around
+                     then Not (seqToStmt tstStmt') `maybePlace` pos:elsStmts'
                      -- otherwise, the cond must fail and wont bind anything
                      else elsStmts'
         return (stmts', assigned3, tmpCount3)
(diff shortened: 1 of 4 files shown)


src/Types.hs
```diff
@@ -2044,7 +2044,16 @@ modecheckStmt m name defPos assigned detism tmpCount final
       else if mustFail assigned1 -- is condition guaranteed to fail?
       then do
         logTyped $ "Assigned by failing conditional: " ++ show assigned3
-        return (tstStmt'++elsStmts', assigned3, tmpCount3)
+        bindings <- mapFromUnivSetM ultimateVarType Set.empty
+                    $ bindingVars assigned3
+        impurity <- lift $ stmtsImpurity tstStmt'
+        let stmts' = if impurity > Pure 
+                     -- if the test is non-pure, need to keep the cond around
```
pschachte, line 2051: You don't actually need to keep the `Cond` around, making you keep the then branch, because you know the condition is going to fail.  You could just turn it into `Not tstStmt'  :  elsStmts'`.

src/AST.hs
```diff
@@ -2902,6 +2902,39 @@ seqToStmt [] = Unplaced $ TestBool
 seqToStmt [stmt] = stmt
 seqToStmt stmts = Unplaced $ And stmts
 
+-- | Find the Impurity of a sequence of Stmts, the maximum impurity of all
+-- statements
+stmtsImpurity :: [Placed Stmt] -> Compiler Impurity 
+stmtsImpurity stmts = List.foldl max PromisedPure
```
pschachte, line 2908: You could just use `maximum`.

pschachte (changes requested):

src/AST.hs
jimbxb: `maximum` fails on the empty list, so I opted for this

src/Types.hs
jimbxb: Good point!

src/AST.hs
pschachte: Ah!  Good call.

pschachte:
Thanks for that!

PR #260, opened 2022-03-08
Stdlib additions

jimbxb:
* added more documentation
* added assignment operators (e.g. `+=`)
* added three-way comparison ("spaceship") operators (`<=>`)
* added integer bit manipulation operators (e.g. `<<`) and respective assignment operators (e.g. `<<=`)

### DIFF
--- a/wybelibs/wybe/int.wybe
+++ b/wybelibs/wybe/int.wybe
@@ -1,3 +1,6 @@
+# purpose: Standard Integer Library
+# since  : 0.1
+
 pragma no_standard_library  # Standard library can't depend on itself!
 
 use foreign library m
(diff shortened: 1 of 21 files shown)


wybelibs/wybe/int.wybe
```diff
@@ -1,31 +1,108 @@
...
+pub def (!x:_ %= y:_) { ?x = x % y }
+
+# Bit-shift left.
+pub def (x:_ << s:_):_ = foreign llvm shl(x, s)
+
+# Bit-shift left assignment.
+pub def (!x:_ <<= s:_) { ?x = x << s }
+
+# Logical bit-shift right (pad with zeroes).
+pub def (x:_ >> s:_):_ = foreign llvm lshr(x, s)
```
pschachte, line 62: For signed integers, I think the "right" version of right shift is the arithmetic one, so that's what the `>>` operator should do.  Having a separate logical right shift is useful, but there probably doesn't need to be a second arithmetic right shift operator.  Also, can you document some intuition for the `>>=` operator?

wybelibs/wybe/int.wybe
```diff
@@ -1,31 +1,108 @@
...
-pub def (?x:_ - y:_):_ = ?z where { ?x = foreign llvm add(z,y) }
-pub def (x:_ - ?y:_):_ = ?z where { ?y = foreign llvm sub(z,x) }
-pub def -(x:_):_ = foreign llvm sub(0,x)
+
+## Common integer operations
+
+# Addition.  This operation is reversible:  you can compute either argument from
+# the other argument and the value.
+pub def (x:_  + y:_):_  = foreign llvm add(x,y)
+pub def (?x:_ + y:_):_  = ?z where ?x = foreign llvm sub(z,y)
```
pschachte, line 17: For consistency, can you wrap the body in braces?.  I'm didn't realise it would compile without them!

pschachte (changes requested):
Thanks for doing this.  Just a couple of nits to be picked.

pschachte (approved):

Issue #263, opened 2022-03-24
Breaks in Use blocks in Do blocks can cause global variables to be in an invalid state


jimbxb:
e.g.
```
resource counter:int = 10

do {
    use counter in {
        ?counter = 1
        while counter = 10
    }
}

!println(counter)
```

The `counter` global should be restored after the `use` block, but it is not restored because the `break` (from the `while`) is contained within the `use` block.

PR #264, opened 2022-03-24
Correctly restore global variables when breaking inside a `use` block in a loop

jimbxb:
Fixes for bug #263

### DIFF
--- a/test-cases/final-dump/break_in_loop_in_do.exp
+++ b/test-cases/final-dump/break_in_loop_in_do.exp
@@ -1,0 +1,94 @@
+======================================================================
+AFTER EVERYTHING:
+ Module break_in_loop_in_do
+  representation  : (not a type)
+  public submods  : 
+  public resources: 
+  public procs    : break_in_loop_in_do.<0>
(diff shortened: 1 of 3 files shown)


test-cases/final-dump/break_in_loop_in_do.exp
```diff
@@ -0,0 +1,94 @@
...
+ Module break_in_loop_in_do
+  representation  : (not a type)
+  public submods  : 
+  public resources: 
+  public procs    : break_in_loop_in_do.<0>
+  imports         : use wybe
+  resources       : counter: fromList [(break_in_loop_in_do.counter,wybe.int = 10 @break_in_loop_in_do:nn:nn @break_in_loop_in_do:nn:nn)]
+  procs           : 
+
+module top-level code > public {impure} (0 calls)
```
pschachte, line 12: This is a fun test case because the loop has been completely inlined, but can you add a more complex case that doesn't get inlined?

pschachte (approved):
The code looks fine, but can you add another test case?

PR #265, opened 2022-03-27
Improved common subexpr optimisation; minor tweaks

jimbxb:
In my own musings, I decided to see if I had correctly performed the common subexpression optimisation correctly for higher order. To my surprise, I think that the current common subexpression optimisation isnt done as well as it could be--the `CallSiteID` is something that should be canonicalised away, as it is not relevant in finding a common subexpression.

This PR fixes that. Test cases have been updated to reflect these changes, and it shows some cases where this optimisation wasnt done.

Minor tweaks are also in the PR, to flesh out the logging messages and attach more position information to the CSE moves.

### DIFF
--- a/src/BodyBuilder.hs
+++ b/src/BodyBuilder.hs
@@ -23,6 +23,7 @@ import UnivSet as USet
 import Data.Maybe
 import Data.Tuple.HT (mapFst)
 import Data.Bits
+import Data.Function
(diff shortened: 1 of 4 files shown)


pschachte (approved):
Thanks for that.  I didn't think of that when CallSiteIDs were added.

Issue #266, opened 2022-03-28
--no-llvm-opt flag is never used


neon64:
As part of working on https://github.com/pschachte/wybe/issues/109, I've found it useful to sometimes dump LLVM IR to a file.
By default however, I believe `wybemk` writes optimized LLVM IR. Although the `--no-llvm-opt` flag exists, it doesn't control anything (searching the code: https://github.com/pschachte/wybe/search?q=optNoLLVMOpt)

I've got a quick fix on my current dev branch: https://github.com/neon64/wybe/blob/d94021a424706c43ddadbc418f443676381f0b36/src/Emit.hs#L112-L121. Hopefully tomorrow or so I can make a separate branch just containing that fix, and submit a PR.

jimbxb:
Hi Chris!

Perhaps we should use `withModule` instead of `withContext` as this would reduce some code duplication. 

Unless you want to force it to not verify the LLVM, which you could just pass in `False` for if that's a requirement?

PR #267, opened 2022-03-30
respect --no-llvm-opt flag

neon64:
Closes https://github.com/pschachte/wybe/issues/266

I didn't see `withModule` originally, so have now refactored to make use of it.

Side note: it looks like my editor likes to automatically trim trailing whitespace. Before I put up a PR to trim trailing whitespace in all the other source files too, is there some significance to the current trailing whitespace that I'm missing?


jimbxb:
On the side note:

There isn't any significance. It's likely just remnants from various people (like me) not being careful about it. I think a linter check might be nice to have overall with the repo, which would require consistent formatting throughout modules which is nice

pschachte (approved):
All good, thanks.

PR #270, opened 2022-03-31
Add -O flag

### LINKED ISSUE #269: Add compiler flags to control the LLVM compiler optimisations
As the title says.

This really comes in two stages
* **COMPLETED** `-O` 
  * Handle one flag, `-O` which is like that of, e,g., GHC, GCC, where you control a curated set of passes
  * LLVM has such curated pass sets, so this is the simplest stage to support
* -`f`
  * `-f` flags, e.g., `-ffast-math`, are much more general, being switches for each particular optimisation
  * this one seems much harder to do
  
 Another point of note is to see if this `-O0` flag would be equivalent to the `--no-llvm-opt` flag (#266)

jimbxb:
Partially closes #269

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -55,6 +55,22 @@ Note that `wybemk` is like `make` in that you give it the name of the
 file you want it to build, and it figures out what files it needs
 to compile.
 
+### Compiler Options
+
+`wybemk` supports various compiler options. A full listing of available flags
+can be found with the following:
+
+```
+% wybemk --help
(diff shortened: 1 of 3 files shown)


pschachte (changes requested):
This looks good, and it's fine to leave out the second part (maybe post an issue).  But could you please document the -O switch in `WYBE.md` file?

pschachte (approved):
Thanks!

PR #271, opened 2022-04-04
remove trailing whitespace + rerun make src/README.md

neon64:
I'm hoping this might make future PRs simpler by getting rid of inconsistent trailing whitespace.

To create this diff, I went and ran "trim trailing whitespace" in my editor for each file in `src/`.

I wonder for the future, whether adopting some sort of autoformatter / linter might make this easier? I've used `hlint` before, but I don't think it does whitespace/formatting stuff.

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -66,8 +66,8 @@ can be found with the following:
 
 #### Optimisation Options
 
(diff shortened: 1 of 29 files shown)


pschachte (approved):
Thanks for that, Chris.

Could you add an issue in the issue tracker asking for a linter?  That's a good idea.

PR #273, opened 2022-04-04
Infer more LLVM types during codegen, remove some unused code

neon64:
I haven't had a chance to discuss this one yet, so I'd be open to any feedback (maybe there's something I'm missing why it's done the way it is?)

- When implementing new LPVM instructions, I accidentally introduced a couple of bugs by falsely claiming an operand was of type `x` (by writing `assign name op x`), when the operand was actually of a different type `y`.
- I learned that each LLVM `Operand` has a type, which can be retrieved with `typeOf operand`

Thus, I thought it might make sense to use the LLVM type of the operand itself, rather than passing in an extra argument. This also simplifies `doCast` and `consCast`.

I believe this code change shouldn't break any of the tests / final dumps. I'd be keen to hear what you think.

I also have an extra commit https://github.com/neon64/wybe/commit/dacb26958184579b1df9d221864e6118c80ba0c6 which removes all the `getelementptr i64, i64* %xyz, i64 0` instructions from `gcAccess` and `gcMutate` which afaik are just nops (I guess they get optimized out by LLVM pretty quickly, but it seemed a bit redundant to generate them in the first place). However, this change breaks nearly every single final-dump test, so I thought I'd keep it out of this PR for now.

### DIFF
--- a/src/Codegen.hs
+++ b/src/Codegen.hs
@@ -690,6 +690,11 @@ csext :: C.Constant -> LLVMAST.Type -> Codegen C.Constant
 csext op ty = return $ C.SExt op ty
 
 
+-- Helpers for allocating, storing, loading
+doAlloca :: Type -> Codegen Operand
+doAlloca (PointerType ty _) = instr (ptr_t ty) $ Alloca ty Nothing 0 []
+doAlloca ty                 = instr (ptr_t ty) $ Alloca ty Nothing 0 []
+
 doLoad :: Type -> Operand -> Codegen Operand
 doLoad ty ptr = instr ty $ Load False ptr Nothing 0 []
 
--- a/src/Codegen.hs
+++ b/src/Codegen.hs
@@ -19,8 +19,7 @@ module Codegen (
   phi, br, cbr, getBlock, retNothing, fresh,
   -- * Symbol storage
   alloca, store, local, assign, addOperand, load, getVar, localVar, preservingSymtab,
-  makeGlobalResourceVariable,
-  operandType, doAlloca, doLoad,
+  makeGlobalResourceVariable, doLoad,
   bitcast, cbitcast, inttoptr, cinttoptr, ptrtoint, cptrtoint,
   trunc, ctrunc, zext, czext, sext, csext,
   -- * Types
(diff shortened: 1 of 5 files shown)


test-cases/final-dump-test.sh
```diff
@@ -23,18 +28,18 @@ do
             -e 's/\[[0-9][0-9]* x i8\]/[?? x i8]/g' \
         > $out
 	# Add a newline to the end of a file if there isn't to resolve platform differences.
-	ed -s $out <<< w > /dev/null 2>&1
-	if [ ! -r $exp ] ; then 
+	ed -s $out <<< w > /dev/null 2>&1 || abort "failed to add newline to output, is ed installed?"
```
neon64, line 31: I realised that this is why the tests were failing locally last week - I didn't have `ed` installed so it silently failed to add newlines.

I added this as wasn't sure if someone in the future might encounter the same issue? Alternatively I could add a note in [INSTALL.md](https://github.com/pschachte/wybe/blob/master/INSTALL.md) rather than modifying this script?

src/Codegen.hs
```diff
@@ -167,7 +167,7 @@ data BlockState
 -- Stores assignments and generated operands
 data SymbolTable
      = SymbolTable {
-         stVars :: Map.Map String (Operand,TypeRepresentation)
+         stVars :: Map.Map String (Operand,LLVM.AST.Type)
```
jimbxb, line 170: Do we still need to store the `Type` here? Can't we just use `typeOf` to get the same information?

jimbxb (commented):
Thanks for the help cleaning up the LLVM codegen. It's not the nicest file to trawl through.

I'd actually encountered this myself when dealing with higher order tests (as the Boolean would be an `i64` and LLVM didn't like that without a `trunc`, and this is why the `castVar` function is now called in a fork). A lot of this code was written before the type system even had generics, to my knowledge, so there's probably a fair few legacy decisions that still exist.

test-cases/final-dump-test.sh
jimbxb: What good is a Unix environment if there's no `ed` installed :(

Probably a good thing to note in INSTALL, but only as something required for tests--something a user wouldn't really be doing

src/Codegen.hs
neon64: good point, I've update the PR to remove that

test-cases/final-dump-test.sh
jimbxb: Scouring SO I have found [this](https://unix.stackexchange.com/a/263965):
```
[ -n "$(tail -c1 file)" ] && echo >> file
```
The poster claims this is POSIX compliant, and also works without `ed` :)

src/Codegen.hs
```diff
@@ -707,11 +690,6 @@ csext :: C.Constant -> LLVMAST.Type -> Codegen C.Constant
 csext op ty = return $ C.SExt op ty
 
 
--- Helpers for allocating, storing, loading
```
pschachte, line 710: You should probably keep this.  I'm sure we're going to want it eventually, and you're probably going to need it for your project.

pschachte (changes requested):
Just one change to reverse, otherwise it's all good.

src/Codegen.hs
neon64: no worries, I've added it back in again

pschachte (approved):

Issue #109, opened 2021-02-22
Implement transformation of output arguments to input arguments holding the address at which to store the output


pschachte:
This transformation will allow a data structure intended to store the result of an operation to be constructed before the operation is invoked.  The most important use of this will be to allow more operations to be tail recursive.  For example, append generally cannot be tail recursive, because a new list cell must be constructed after the recursive call:

```
def append(front::list, back::list, ?result::list) {
    if {[?h | ?t] = front ::
        append(t, back, ?tail)
        ?result = [h | tail]
    otherwise::
        ?result = back
    }
}
````

With this transformation, the assignment `?result = [h | tail]` can be performed before the call to append.

For illustrative purposes, I will use the syntax `&x` to denote an output argument implemented as an input, and `?&x` to denote a variable being assigned the address of the member in which that value should be stored.  (This syntax should not be added to the language; I'm only using it to explain the transformation.  This transformation should be transparent for the Wybe programmer.)

With that syntax, the transformation would produce:

```
def append(front::list, back::list, &result::list) {
    if {[?h | ?t] = front ::
        ?result = [h | ?&tail]
        append(t, back, &tail)
    otherwise::
        ?result = back
    }
}
````

After this, a call such as `append(x, y, ?z)` would be transformed to `append(x, y, &z)`, where &z is the address of the local variable `z`.

PR #279, opened 2022-04-18
Implement last call modulo construction optimization

neon64:
The branch I've been playing around with for the last month or so is finally beginning to converge to something usable, so I thought I should finally open this PR to gather some feedback!

Implements an MVP for https://github.com/pschachte/wybe/issues/109

### Missing pieces

- this branch fails a few tests. I have had a look through `ERRS` and I believe all the differences are expected, i.e.: they are a result of this optimization kicking in and therefore affecting the final dump LPVM/LLVM. I'm not sure the best way to easily show/summarize this though.

- I haven't updated [WYBE.md](/pschachte/wybe/blob/master/wybe.md) yet, with e.g.:  details of what how `outByReference` flow should work

### Implementation notes (since last meeting)

The main guts of this PR are probably 50/50 between `src/LastCallAnalysis.hs` for the LPVM transformation, and `src/Blocks.hs` for the codegen changes. For this MVP version I only handle the case of directly recursive functions, and don't hook into multiple-specialization at all.

The last call analysis pass (added to `src/Builder.hs`) currently runs *before* all the specialized versions get generated (in `multiSpeczTopDownPass`). This was the cause of "empty function bodies" bug I mentioned earlier, i.e.: specialized function bodies looking something like this:

```
Blocks: nrev<0>[somemultispeczid](lst##0:generic_list(T),  outByReference #result##0:generic_list(T))<{}; {}>
Blocks:     foreign llvm icmp_ne(lst##0:wybe.int, 0:wybe.int, ?tmp#8##0:wybe.bool)
Blocks:     case ~tmp#8##0:wybe.bool of
Blocks:     0:
Blocks:
Blocks:     1:
```

I did some debugging and I believe it was because the analysis only considered `FlowOut` params as outputs, so it thought the whole function body had no outputs (i.e.: after turning on `-l BodyBuilder`, the list of `outSubsts` was always empty), therefore removed every prim since they "weren't needed". I believe there are two possible solutions for this, either I could move LastCallAnalysis to after specialized versions are generated, or I make the other passes handle post-LastCallAnalysis proc bodies. I tried the latter option, and made a couple of changes to `src/Transform.hs` to treat the `FlowOutByReference` flow the same as a `FlowOut` flow. This nearly worked, except it didn't like the inverted ordering of mutates before the tail call:

```prolog
foreign lpvm mutate(~tmp#10##0:wybe.list(T), ?result##0:wybe.list(wybe.int), 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, takeReference tail##0:wybe.list(T)) @list:8:27
tail append.append<0>[410bae77d3](~t##0:wybe.list(wybe.int), ~back##0:wybe.list(wybe.int), outByReference ~tail##0:wybe.list(wybe.int)) #1 @append:7:13
```

In the specialized body, the final call `tail append.append<0>[410bae77d3]` was simply missing. I believe this was since BodyBuilder saw `outByReference ~tail##0:wybe.list(wybe.int))` as being created right at the end, but never used after (not in `usedLater`). To be honest, this BodyBuilder code was starting to be way out of my depth, and I wasn't sure it was a good idea to fundamentally change how it analyses LPVM. Thus, I instead made a change to `LastCallAnalysis.hs` to keep the final "mutates-that-actually-calculate-addresses" after the tail call, e.g.: 

```prolog
foreign lpvm alloc(16:wybe.int, ?tmp#8##0:generic_list(T)) @generic_list:nn:nn
foreign lpvm mutate(~tmp#8##0:generic_list(T), ?tmp#9##0:generic_list(T), 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~h##0:T) @generic_list:nn:nn
tail generic_list.append<0>(~t##0:generic_list(T), ~y##0:generic_list(T), outByReference tmp#2##0:generic_list(T)) #1 @generic_list:nn:nn
foreign lpvm mutate(~tmp#9##0:generic_list(T), ?#result##0:generic_list(T), 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~takeReference tmp#2##0:generic_list(T)) @generic_list:nn:nn
```

It was then relatively straightforward to modify `src/Blocks.hs` to pick up that any Prims after a call marked with a new `tail` LPVM attribute must be *"mutates-that-actually-calculate-addresses"*, and so calculate the addresses *before* making the call. This also means that the fourth flow, `FlowTakeReference`, is superfluous as one can infer that any `mutate()` after a call marked with `tail` must necessarily have a `takeReference` final arg. I wonder then, is it overkill to introduce an entirely new flow which is only ever used in the last argument of a `mutate()` call? On the other hand, there is a nice symmetry between `outByReference` and `takeReference`.

### Types in codegen

In my first implementation of this optimisation (the "explicit version" - [at this commit](https://github.com/pschachte/wybe/blob/bfa0baa8d563f282de7d14c30a30253d087bd754/src/LastCallAnalysis.hs)), I added new a new `Pointer` `TypeRepresentation` which was used by `outByReference` parameters. I now agree that that isn't a good approach for LPVM, and so now codegen takes a more implicit approach, where it looks at the both the wybe type, and `PrimFlow` of a parameter, in order to decide which LLVM type it should use. If flow is `FlowOutByReference` or `FlowTakeReference`, then actual LLVM type is `ptr_t originalType`.

There is also a concept of a "shadow reference variable". When calling a function with an `outByReference` parameter, if we don't have an existing address to store the output, codegen will instead stack-allocate space for said output, and pass this "shadow reference variable" (`%65` in the example below), and then load the actual value afterwards (into `%66` below).

```
Blocks:   %65 = alloca i64
Blocks:    call fastcc  void  @"append.append<0>[410bae77d3]"(i64  %27, i64  %59, i64*  %65)
Blocks:   %66 = load  i64, i64* %65
```

I call this a "shadow" variable since it is internal to codegen. The LPVM code only cares about the underlying value - i.e.: given the LPVM below, there is no way to access or see the shadow var `x3#ref##0` that codegen creates internally just before the call to append.

```prolog
foreign lpvm mutate(~tmp#40##0:wybe.list(T), ?tmp#41##0:wybe.list(T), 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 5:T) @list:8:27
foreign lpvm mutate(~tmp#41##0:wybe.list(T), ?tmp#5##0:wybe.list(wybe.int), 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, ~tmp#6##0:wybe.list(T)) @list:8:27
append.append<0>[410bae77d3](~tmp#0##0:wybe.list(wybe.int), tmp#5##0:wybe.list(wybe.int), outByReference x3##0:wybe.list(wybe.int)) #10 @append:21:1
wybe.io.print_list<0>(~tmp#5##0:wybe.list(wybe.int))<{<<wybe.io.io>>}; {<<wybe.io.io>>}> #11 @append:23:2
wybe.io.print_list<0>(~x3##0:wybe.list(wybe.int))<{<<wybe.io.io>>}; {<<wybe.io.io>>}> #12 @append:24:2\
```

### Summary

Oops, there is a lot to go through here! I've tried my best to articulate some of the things I've encountered over the last week in writing, but would be more than happy to talk through them in this week's meeting, using this very-long PR description as summary notes.


src/Blocks.hs
```diff
@@ -586,6 +613,40 @@ cgen prim@(PrimForeign lang name flags args) = do
...
+-- | We observe hints from previous stages of compilation (`attrs`), as well
+-- as the callee's signature, in order suggest a stricter LLVM
+-- TailCallKind where possible.
+-- In the best case, the call is marked `musttail`, which signals LLVM must
+-- tail-call-optimize.
+-- `tail` is just a hint that LLVM "can" tail-call optimize but doesn't have to.
+-- However there are certain restrictions on when these hints. LLVM may optimize
+-- our code incorrectly (or crash) if we get this wrong.
+getTailCallHint :: Set PrimCallAttribute -> [PrimArg] -> LLVMAST.Type -> Maybe LLVMAST.TailCallKind
+getTailCallHint attrs args outTy =
```
neon64, line 628: ...and as if my PR description wasn't long enough, I have some more interesting things I discovered this weekend. :smile: 

It turns out just blindly adding `tail` to every LLVM call can trigger some Very Bad Behaviour in LLVM.

For example, consider this `wybemk` output after being passed through LLVM's optimizer:
https://godbolt.org/z/GrcWv65qs

Note that a `load` has disappeared and an `undef` has crept in. That is why my append test program printed `[0]` in `-O3` and `[1,2,3,4,5,6]` in `-O1`. At first I naively thought this could be an LLVM bug. But it turns out LLVM's language reference has this to say about the `tail` attribute:

2022-04-16T23:20:05,323375165+10:00

So in this case, since my optimisation now starts adding `allocas` in some places, we are producing Bad IR<sup>TM</sup> and so LLVM is free to do what it wants.
 
My slightly solution thus far is to be more careful when we mark calls with `tail` (and now also `musttail`) LLVM attributes. My suggested implementation looks at both the presence of the `tail` LPVM attribute (which is effectively a promise that the callee don't use any allocas from the caller*), and also the signature of the callee, to infer a `TailCallKind` which may be slightly conservative, but hopefully won't break LLVM.

*`LastCallAnalysis.hs` is the only place which adds this `tail` LPVM attribute to LPVM calls, and it should guarantee this promise

pschachte: There's a script update-exp in the top-level directory that runs all tests and does a colourised word diff between the actual and expected outputs, and asks whether the new output is what should be expected, and if so, replaces the expected output file with the new one.  The colourised word diff makes it a lot easier to see what's actually different compared to the ERRS file, although it's still laborious.  If you go through all the differences and you can mark all of them correct, and push the changes to the .exp files, then the PR will pass all tests.

If you haven't made the new flows user-visible, then there is probably not much need for documentation changes.  At some point, there should probably be a section toward the end of the doc that talks about efficiency and how to write Wybe code to be as efficient as possible, which would be the place to document your change, but that doesn't exist yet.

There's a function `flowsOut` defined in AST.hs that works for `FlowDirections`, which is the Wybe source code version of `PrimFlow`.  Since it's no longer correct to just pattern match against `FlowOut` to check if a `PrimFlow` is an output, you should define a function `primFlowsOut` in AST.hs that returns `True` for both `FlowOut` and `OutByReference` flows, and use that to decide if a proc produces any outputs.  It's probably worth grepping the code for FlowOut and see if there are any other places that need to use the new `primFlowsOut` function.

OK, it's fair enough to keep the mutate instruction at the end, and somehow mark it to indicate that instead of doing it, the LLVM generation should take the address of the memory it would have mutated.  The tricky thing about doing it that way is that you have to generate LLVM code for the mutate instruction before generating code for the preceding call.

Yes, BodyBuilder does get pretty subtle.  I'm not sure we'll be able to get away with keeping it ignorant of the OutByReference flow in the long run, but we can certainly try.

Nice catch about the LLVM handling of tail calls.  I didn't realise that restriction existed, although on reflection I should have.  Your solution of not marking non-tail calls as tail calls sounds like the right one.

src/LastCallAnalysis.hs
```diff
@@ -0,0 +1,360 @@
...
+updateEachProcM f (AcyclicSCC pspec) = updateProcDefM f pspec
+updateEachProcM f (CyclicSCC pspecs) = mapM_ (updateProcDefM f) pspecs
+
+-- | Analyse whether a proc is suitable for last-call -> tail-call transformation.
+-- If so, we:
+--   (1) modify the function's signature, switching some outputs to FlowOutByReference
+--   (2) mark the last call with the LPVM `tail` attribute
+--   (3) mark the mutate() instructions after the last call with `takeReference` flows
+lastCallAnalyseProc :: ProcDef -> Compiler ProcDef
+lastCallAnalyseProc def = do
```
jimbxb, line 52: When I introduced closures, I added a field in `ProcDef`s that dictaty the variant of a procedure, which says if the proc is a constructor, deconstructor, etc. The really important one is the `Closure` variant, that is applied in the case the procedure originates from a closure in the code--either a partial application of some other proc or an anonymous procedure. 

I think you should not apply this analysis to such Closure procedures. These closure procedures have a very particular interface in the LLVM generation, and this analysis may violate that. I am unsure if this can occur, but a defensive check is quite cheap and simple to implement, so should be no harm

jimbxb (changes requested):

src/LastCallAnalysis.hs
pschachte: It may be necessary to do that transformation on closure procedures.  If the proc you're closing over has an out-by-reference argument, it's necessary to allocate a stack slot, pass its address to the proc, and then fetch the value out of the stack slot.  You're right that we don't want to transform a closure proc to change its parameter passing, but for the moment we only do that for recursive procs, and closure procs won't be recursive, so that shouldn't be a problem.  But it couldn't hurt to put in a test to forever rule out changing the flow direction of a closure proc.

neon64: thanks for looking at this @jimbxb , it's been on my todo list to see how this optimization interacts with higher-order procs. Maybe I will try and find a test-case where this could be a problem

test-cases/final-dump/higher_order_append.exp
```diff
@@ -0,0 +1,247 @@
...
+
+define external fastcc  i64 @"higher_order_append.gen#1<1>"(i64  %"#env##0", i64  %"anon#1#1##0")    {
+entry:
+  %89 = add   i64 %"#env##0", 8 
+  %90 = inttoptr i64 %89 to i64* 
+  %91 = getelementptr  i64, i64* %90, i64 0 
+  %92 = load  i64, i64* %91 
+  %93 = alloca i64 
+   call fastcc  void  @"higher_order_append.append<0>"(i64  %"anon#1#1##0", i64  %92, i64*  %93)  
+  %94 = load  i64, i64* %93
```
neon64, line 245: I have added this test-case to ensure that we call the last-call-optimised version of functions (e.g.: append) inside of closure procs (note the `alloca` and then `load` immediately after)

src/LastCallAnalysis.hs
```diff
@@ -55,13 +55,18 @@ lastCallAnalyseProc def = do
...
-        LeafTransformState { procSpec = spec, originalProto = proto, transformedProto = Nothing }
-    case (maybeBody', transformedProto finalState) of
-        (Just body', Just proto') -> do
-            logLastCallAnalysis $ "new proto: " ++ show proto'
-            return def {procImpln = impln{procImplnProto = proto', procImplnBody = body'}}
-        _ -> return def
+    case procVariant def of
+        ClosureProc {} -> do
+            logLastCallAnalysis "skipping closure proc, shouldn't have direct tail-recursion anyway"
+            return def
```
neon64, line 61: have explicitly skipped transforming closure procs for now, even though they should never be directly tail recursive so should never be eligible for transformation

test-cases/final-dump/higher_order_append.exp
jimbxb: Great. Looks okay by me

wybelibs/wybe/list.wybe
```diff
@@ -4,6 +4,7 @@
 pragma no_standard_library  # Standard library can't depend on itself!
 
 use wybe.bool, wybe.int
+use wybe.io, wybe.string, wybe.c_string, wybe.char, wybe.range, wybe.phantom, wybe.count, wybe.float
```
neon64, line 7: I had some trouble implementing `print` for lists and lists of lists, but am not sure if its a compiler bug or just me being silly. First problem: the compiler complained these types were unknown, until I added all of these imports

```
./wybemk --force-all -L wybelibs wybelibs/wybe.o
Error detected during type checking of module(s) wybe, wybe.array, wybe.array.raw_array, wybe.control, wybe.string, wybe.io, wybe.list, wybe.range, wybe.machine_word
In proc declaration, unknown type wybe.char
In proc declaration, unknown type wybe.phantom
In proc declaration, unknown type wybe.count
In proc declaration, unknown type wybe.float
In proc declaration, unknown type wybe.string
make: *** [Makefile:52: wybelibs/wybe.o] Error 1
```

wybelibs/wybe/list.wybe
```diff
@@ -75,3 +76,38 @@ pub def filter(p:{test}(A), !as:_(A)) {
...
     ?as = reverse(out)
 }
+
+pub def print(x:list(int)) use !io {
+    !print("[")
+    !_print_list(x)
+    !print("]")
+}
+
+pub def print(xs:list(list(int))) use !io {
```
neon64, line 86: The compiler also wasn't happy when I had this overload (for `list(list(int))`) in the `wybe.io` module. I saw https://github.com/pschachte/wybe/pull/287/files though, so I have just put these in `wybe.list` from the get-go for now.

```
./wybemk --force-all -L wybelibs wybelibs/wybe.o
Error detected during type checking of module(s) wybe, wybe.array, wybe.array.raw_array, wybe.control, wybe.string, wybe.io, wybe.range, wybe.machine_word
In proc declaration, unknown type wybe.list(wybe.int)
In proc declaration, unknown type wybe.list(wybe.list(wybe.int))
make: *** [Makefile:52: wybelibs/wybe.o] Error 1
```

wybelibs/wybe/list.wybe
jimbxb: This occurs because of the `pragma no_standard_library`. You need to add all of the imports for each of the types as theyre not public imports in `io`... I think this is a bug tbh

wybelibs/wybe/list.wybe
jimbxb: Same deal here. `io` doesnt know about `wybe.list`--it shouldnt need to

neon64: To clarify this one: I did add "use wybe.list" at the top of io.wybe and it still didn't work. But it only started complaining once I added the "list(list(int))" overload ("list(int)" was working fine as of a few commits ago). To me it seems like the typechecker doesn't like how the type parameter refers to the type itself? (i.e.: T=list(T'), T' = int)

pschachte: Please, when you have time, try to replicate this in a minimal test case and file a bug report.

pschachte (changes requested):
This looks good, except that I merged another PR that conflicts with this.  That was James's PR moving the print procs into the modules for what they were printing.  Unfortunately, that make widespread but trivial changes to the .exp files.  Could you please update this PR to fix the merge conflicts and let me know when it's ready, and I'll merge it?

wybelibs/wybe/io.wybe
```diff
@@ -7,6 +7,7 @@ use foreign object cbits
 use foreign library gc
 
 use wybe.bool, wybe.int, wybe.count, wybe.float, wybe.char, wybe.string, wybe.c_string
+use wybe.list
```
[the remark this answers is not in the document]
neon64: ah no it isn't, looks like I made a commit locally to remove it, but forgot to push the branch. Will update

wybelibs/wybe/list.wybe
```diff
@@ -75,3 +76,38 @@ pub def filter(p:{test}(A), !as:_(A)) {
     }
     ?as = reverse(out)
 }
+
+pub def print(x:list(int)) use !io {
```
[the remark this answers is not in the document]
neon64: okay I've attempted that in https://github.com/pschachte/wybe/pull/279/commits/8ba5099fcf91cae8cf38c53f4b6785e49462e388#diff-5430fa35f7b4236a1975f749ce16b553cae9ea6fad20736b2c73529d27813f7f

neon64:
I merged in latest changes and got tests passing again. And I've put on my todo list to replicate/file a bug report for https://github.com/pschachte/wybe/pull/279#discussion_r867558363

pschachte (approved):

PR #283, opened 2022-04-27
Improved backwards analysis with global flows

jimbxb:
As the title says. This strengthens the definition of neededness in terms of global flows
i.e., assuming a prim is pure and has no regular *needed* outputs, a prim is needed iff the outwards global flows of the prim are not overwritten by some later prim

### E.g.1:
```
resource res:int

def {noinline} out use ?res { ?res = 11 }

use res in {
    ?res = 10
    !out
    !println(res)
}
```

The `?res = 10` assignment is redundant

### E.g.2: 
```
resource res:int

def {noinline} out use ?res { ?res = 11 }

use res in { 
    !out
    ?res = 10 
    !println(res)
}
```
No global flows for `res` occur in the top-level, as the `!out` call can now be omitted. 
This applies to all prims where the outwards global flows are overwritten later on.

### DIFF
--- a/test-cases/final-dump/out_only_global_flow.exp
+++ b/test-cases/final-dump/out_only_global_flow.exp
@@ -1,0 +1,73 @@
+======================================================================
+AFTER EVERYTHING:
+ Module out_only_global_flow
+  representation  : (not a type)
+  public submods  : 
+  public resources: 
+  public procs    : out_only_global_flow.<0>
+  imports         : use wybe
+  resources       : res: fromList [(out_only_global_flow.res,wybe.int @out_only_global_flow:nn:nn)]
+  procs           : 
+
+module top-level code > public {impure} (0 calls)
+0: out_only_global_flow.<0>
+()<{<<wybe.io.io>>}; {<<wybe.io.io>>}>:
+  AliasPairs: []
+  InterestingCallProperties: []
+    foreign lpvm load(<<out_only_global_flow.res>>:wybe.int, ?%tmp#0##0:wybe.int) @out_only_global_flow:nn:nn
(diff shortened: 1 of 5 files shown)


jimbxb:
Closing for now. I want to handle another case regarding neededness.

jimbxb:
Re-opening -- see updated description

test-cases/final-dump/out_only_global_flow.exp
```diff
@@ -0,0 +1,73 @@
...
+  imports         : use wybe
+  resources       : res: fromList [(out_only_global_flow.res,wybe.int @out_only_global_flow:nn:nn)]
+  procs           : 
+
+module top-level code > public {impure} (0 calls)
+0: out_only_global_flow.<0>
+()<{<<wybe.io.io>>}; {<<wybe.io.io>>}>:
+  AliasPairs: []
+  InterestingCallProperties: []
+    foreign lpvm load(<<out_only_global_flow.res>>:wybe.int, ?%tmp#0##0:wybe.int) @out_only_global_flow:nn:nn
```
pschachte, line 17: This load looks unnecessary.  Shouldn't it have been eliminated?

jimbxb: The reason this load cannot be removed is because of the outwards global flow in the call to `out`. We cannot be certain that the global variable is unchanged here, so we must `load` it and subsequently re`store` it, following the `use` block.

If we could inline `out` then we can actually get rid of this `load` and subsequent re`store`, much like in the other test case. This isn't possible though, as the `out` procedure is intentionally marked with `noinline`, and we cannot guarantee the state of the global variable remains unchanged.

pschachte: Aha, I missed that the store at the end used that loaded value.  Fair enough.

pschachte (approved):

PR #290, opened 2022-05-10
Restart virtual register count from 0 for each proc

### LINKED ISSUE #288: LLVM pretty-printed by `ASTShow.hs` differs from LLVM written by `Emit.hs`
Just noticed yesterday that copy-pasting the LLVM printed in the final-dump test cases didn't compile in godbolt, e.g.: https://godbolt.org/z/714bo13zz

It seems like the main difference is that the actual generated LLVM code (`type_generics.ll`) restarts temporaries at `%0` at the start of every function, whereas the pretty-printed LLVM (`type_generics.exp`) starts at `%1` and keeps increasing throughout the whole file.

Here's a brief example of the slight differences:
image

The language reference says the following (which backs up the compile errors in godbolt) 
> Unnamed temporaries are numbered sequentially (using a per-function incrementing counter, starting with 0). Note that basic blocks and unnamed function parameters are included in this numbering. For example, if the entry basic block is not given a label name and all function parameters are named, then it will get number 0.

I guess it doesn't matter if the FinalDump LLVM is accepted by LLVM or not, but it seems a little fishy that it should be so different from the LLVM we actually produce. I think the reason for the difference is that `ASTShow.hs` uses https://github.com/llvm-hs/llvm-hs-pretty to do LLVM IR pretty-printing, whereas `Emit.hs` uses `writeLLVMAssemblyToFile` which uses the actual LLVM implementation. I've tested fixing `ASTShow.hs` by switching to use `LLVM.moduleLLVMAssembly` instead. The downside is that it requires an LLVM context (thus inside `IO` monad), thus `Module` cannot implement `Show` anymore.

One benefit to changing the final dump LLVM IR to match the emitted `.ll` files (apart from consistency), is that since unnamed temporaries would restart at %0 each function, the changes to final-dump outputs would become less noisy, e.g.: this example where some of the diff is just renaming from `%70` to `%71` and onwards, obscuring the real changes.

image

jimbxb:
Closes #288 

@neon64, is this sufficient for a fix?

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -39,8 +39,8 @@ module AST (
   lookupTypeRepresentation, lookupModuleRepresentation,
   paramIsPhantom, argIsPhantom, typeIsPhantom, repIsPhantom,
   primProtoParamNames,
-  protoRealParams, realParams, paramIsReal,
-  protoInputParamNames, isProcProtoArg,
+  protoRealParams, realParams, paramIsReal, paramIsNeeded,
+  protoInputParamNames, protoRealParamsWhere, isProcProtoArg,
   -- *Source Position Types
   OptPos, Placed(..), place, betterPlace, content, maybePlace, rePlace, unPlace,
   placedApply, defaultPlacedApply, placedApplyM, contentApply, updatePlacedM,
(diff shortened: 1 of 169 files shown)


neon64:
Ah great work! While the emitted code is still different in some ways to the code emitted in a `.ll` file, it does seem like a sufficient fix insofar that the emitted LLVM IR at least now works when fed into `llc` or `opt` (at least from my n=1 experiment: https://godbolt.org/z/edn98Kjsr), which was the main thing I was keen for.

In the below screenshot (taken from #288), you can see there are other small changes aside from the numbering starting from `%0`. But as you mentioned in https://github.com/pschachte/wybe/issues/288#issuecomment-1119257271, there are some downsides to switching to `LLVM.moduleLLVMAssembly` in order to perfectly replicate the `.ll` output. This compromise seems pretty good to me

jimbxb:
I believe some of the information, such as `preds` wouldn't be hard to fill in--LPVM had some very nice properties about what the predecessor of each basic block will be :)

pschachte (changes requested):
Sorry, James, the change looks good but I've merged a couple of PRs, so now many of the .exp files conflict.  Could you please refresh them?

jimbxb:
No worries. On a fix now.

jimbxb:
I also removed the `namedInstr` function. This is to unify the way that returns are handled--there was no need for the LLVM to use names here, instead now everything is unnamed

Issue #300, opened 2022-05-16
Explicit imports are required for types that appear in a different, unrelated proc overload


neon64:
I finally got around to investigating this [compile error](https://github.com/pschachte/wybe/pull/279#discussion_r867558363_) which came up when trying to modify the standard library to add more `print` overloads.

I have a pretty minimal test case which demonstrates what goes wrong:

```
Error detected during type checking of module(s) overloaded_print.mybool, overloaded_print.myio, overloaded_print.mylist
final-dump/overloaded_print/mybool.wybe:9:37: In proc declaration, unknown type overloaded_print.mylist(wybe.int)
```

The reason why I believe this is a compiler bug, is that `mybool.wybe` never refers to the type `mylist`, so I believe shouldn't need to import it. However there is a different overload of `myprint` (defined in the module `myio`) which *does* refer to `mylist`. I believe the typechecker is getting confused between the two .

Here is the definition of [mybool.wybe](https://github.com/neon64/wybe/blob/typechecking_error/test-cases/final-dump/overloaded_print/mybool.wybe) and [myio.wybe](https://github.com/neon64/wybe/blob/typechecking_error/test-cases/final-dump/overloaded_print/myio.wybe)

I haven't looked to much into the code implementing Wybe's typechecker yet. Maybe with this minimal test-case someone can point me to an easy fix, otherwise I will look into it myself more tomorrow

neon64:
Thought I should include an excerpt from the compiler log output (sorry it's very verbose):

It seems like we call `lookupType` on the "incorrect" overload which ultimately triggers the compilation error:
 `AST: In module overloaded_print.mybool, looking up type overloaded_print.mylist(wybe.int)`

```
Types: Unifying types wybe.char (-> wybe.char) and wybe.bool (-> wybe.bool)
Types: Recording error(s): [Error Just "/home/chris/code/wybe/test-cases/final-dump/overloaded_print/mybool.wybe" (line 9, column 37): Type error in call to myprint, argument 1]
Types:   Unification yields XXX
Types: Matching types [wybe.char] with {impure} overloaded_print.myio.myprint<0>(wybe.char)
AST: In module overloaded_print.mybool, looking up type wybe.char
AST: Finding visible symbol wybe.char from module overloaded_print.mybool
AST: *** ALL matching visible modules: wybe
AST: Candidates: wybe.char
AST: Get loaded module wybe.char
AST:  worked
AST: Matching type = wybe.char
[...]

PR #303, opened 2022-05-20
Fix for #300

jimbxb:
Regarding the new test case, I believe in the future we should raise an error if there is a public proc with a parameter whose type is not public.

The test case has such, with `hidden`


src/Types.hs
```diff
@@ -540,20 +544,24 @@ instance Show Typing where
        then " (with no errors)"
        else " with errors: " ++ show errs
 
+lookupTyped :: String -> OptPos -> TypeSpec -> Typed TypeSpec
```
pschachte, line 547: Could you add a bit of doc?  The name is a bit mysterious to me.

test-cases/final-dump/unexported_type_overload.wybe
```diff
@@ -0,0 +1,9 @@
+module sub {
+    pub def proc(x:char) { pass }
+
+    pub def proc(x:hidden) { pass }
```
pschachte, line 4: I thought this PR was meant to give an error message in this case?

pschachte (changes requested):

test-cases/final-dump/unexported_type_overload.wybe
jimbxb: Yeah, so all this PR does is remove a spurious error.

As there is a correct overload (the `char` one), all is okay with no error.

I can add a test case to show the error is reported if the overload fails

jimbxb: Nevermind, I found a way. Its a hack and a half around the type system though..

pschachte (approved):

PR #304, opened 2022-05-20
add benchmarking library

neon64:
for my presentation+report I was planning to include some benchmarks, I think we briefly mentioned that it could be advantageous to add some benchmarking helper functions.

Here's an example of using the library:

```
use benchmark

def naive_reverse(x: list(T), ?y: list(T)) {
    if {
        [ ?h | ?t] = x ::
            naive_reverse(t, ?t2)
            ?y = t2 ,, [h]
        | else ::
            ?y = []
    }
}

def build_list_of_length(n: int, x0: int, ?l: list(int)) {
    # implementation omitted
}

build_list_of_length(200000, 0, ?l)
!benchmark.measure({ naive_reverse(@1, ?@2) }, l, ?l2, ?time)
!println(l2, int.print)
!print("benchmark took: ")
!println(time)
```

The current definition for `measure` is

```
pub def measure(func: (A, ?B), input: A, ?output: B, ?seconds_elapsed: float) use !io
```

I'm open to any suggestions on how to improve the interface. `use !io` is just a marker here since I'm calling an external C `clock()` function, it doesn't write/read anything from stdout/stdin. There are also two low-level primitives `start` and `finish`, however they have the disadvantage that failing to call `finish()` will leak memory, although I did discover the `{unique}` type annotation which should prevent double frees.


test-cases/execution/benchmark_example.wybe
```diff
@@ -0,0 +1,31 @@
...
+    if { n = 0 ::
+        ?l = []
+        | else ::
+        build_list_of_length(n-1, x0 + 1, ?t)
+        ?l = [x0 | t]
+    }
+}
+
+build_list_of_length(100, 0, ?l)
+!println(l, int.print)
```
jimbxb, line 23: Is there a reason the higher-order proc is the second argument?

I think it makes sense as the first, as it allows for partial application without the use of an anon proc

wybelibs/benchmark_impl.c
```diff
@@ -0,0 +1,27 @@
...
+#include <stdbool.h>
+#include <stdlib.h>
+#include <assert.h>
+
+typedef struct {
+    clock_t start;
+} benchmark_state_t;
+
+benchmark_state_t *benchmark_start() {
+    benchmark_state_t *benchmark_state = (benchmark_state_t*) malloc(sizeof(benchmark_state_t));
```
jimbxb, line 15: Im curious what the C standard says about the size of `clock_t`. If its within a word size, do we need to malloc? We should just be able to just cast it to a `void*` if that's the case.

If we can remove the `malloc` and `free`, I think that would be beneficial.

jimbxb (changes requested):

wybelibs/benchmark_impl.c
neon64: I agree it would be nice to get rid of the malloc and free. https://en.cppreference.com/w/c/chrono/clock_t it says it has "implementation defined range and precision" - on my machine `sizeof(clock_t)` is 8, and I imagine that would be the case for most implementations, but it seems like the standard doesn't specify any upper-bound

test-cases/execution/benchmark_example.wybe
neon64: Ah yes this was in a previous PR I believe. To answer your original question though, there wasn't a particular reason I did it that way, so more than happy to swap them

Also you're right, `int.` is redundant, I've updated the PR

wybelibs/benchmark_impl.c
jimbxb: Annoying. I suppose this will do then

test-cases/execution/benchmark_example.wybe
jimbxb: Glad the `int.` is redundant. Otherwise I would have a bug to fix...

jimbxb (approved):

wybelibs/benchmark_impl.c
jimbxb: Actually, I'm not so sure. It seems that it is implementation dependent, but in most implementations it's never longer than 8 words.

@pschachte thoughts?

pschachte: Hmmm.  Is it permissible in C for a function to return something larger than a word?  If not, then it should be OK to assume clock_t is at most word size.

pschachte: If we do need a malloc, it would be better to use GC_malloc, so we don't need to link in two allocators.

pschachte: Better still, you don't really need to use a static variable to hold onto the start clock time.  You could just have one function to get the CPU clock time in seconds as a float, and do the subtraction on the Wybe side.  Then the get_clock_time C function could use alloca() to allocate (and automatically free) the memory.

test-cases/execution/benchmark_example.wybe
```diff
@@ -0,0 +1,31 @@
...
+        ?l = []
+        | else ::
+        build_list_of_length(n-1, x0 + 1, ?t)
+        ?l = [x0 | t]
+    }
+}
+
+build_list_of_length(100, 0, ?l)
+!println(l, print)
+!benchmark.measure({ naive_reverse(@1, ?@2) }, l, ?l2, ?time)
```
pschachte, line 24: I think it'd be nicer is the call to benchmark.measure just took a fully saturated closure (ie, all argument supplied) as argument.  Eg, !benchmark.measure(naive_reverse(l,?l2), ?time).

@jimbxb: if measure is declared impure, will that stop the compiler from removing the call to the closure argument if its output is not used?  That'd be desirable, so you don't have to fight the optimiser to get your code to run.

@neon64: if that works, then you wouldn't need to thread the io state.  I figure this library is just for benchmarking, so purity isn't really important.

jimbxb: `measure` being declared `impure` will force calls to `measure` to be made. You'd need to declare the closure as `impure` to ensure it isn't removed.

neon64: Thank you both for all the comments!

> I think it'd be nicer is the call to benchmark.measure just took a fully saturated closure (ie, all argument supplied) as argument. Eg, !benchmark.measure(naive_reverse(l,?l2), ?time).

I tried this out, didn't have much luck with `!benchmark.measure(naive_reverse(l,?l2), ?time)` (I think since it's trying to pass a second non-existent output argument of  `naive_reverse(l,?l2)` to `measure`)

```
test-cases/execution/benchmark_example.wybe:24:19: Call from module top-level code to naive_reverse with 3 argument(s), expected 2
```

Also had trouble with passing an impure closure:
```
build_list_of_length(100, 0, ?l)
!println(l, print)
benchmark.measure({impure} { naive_reverse(l, ?l2) }, ?time)
!println(l2, print)
if { time > 0.0 && time < 0.01 ::
    !println("benchmark took a reasonable amount of time")
    | else ::
    !println("benchmark took too long: ")
    !print(time)
}
```

```test-cases/execution/benchmark_example.wybe:25:2: Uninitialised argument in call to println, argument 1```

here I think the compiler is right  - we have no guarantee what measure actually does with the closure that we pass in, so it could be that the closure is never called and `l2` is indeed uninitialised by the time we get to the `println`

File: wybelibs/benchmark_impl.c (hunk 2 above)
Comment (neon64) [reply wybelibs/benchmark_impl.c:15]: I believe it is possible for a C function to return something larger than a word (e.g.: returning an arbitrary struct is allowed).

> Better still, you don't really need to use a static variable to hold onto the start clock time. You could just have one function to get the CPU clock time in seconds as a float, and do the subtraction on the Wybe side.

Thanks, I think this should do it (will push a commit once I work out the impure closure issue):

```
double get_clock_time() {
    return ((double) clock()) / CLOCKS_PER_SEC;
}
```

Comment (jimbxb) [reply wybelibs/benchmark_impl.c:15]: I'm going to hazard a guess and say that the `clock_t` type can fit into a word. It is specified by the C standard to be an arithmetic type, which leads me to believe it would be an integer type, as casting it to double is okay then.

But I reckon it is "cleaner" to do the division on the C side of things. We do want to lessen the reliance on C as much as possible, imo

Comment (pschachte) [reply wybelibs/benchmark_impl.c:15]: Now marking a proc as `impure` or `semipure` will stop unneeded calls in the body from being optimised away, or repeated calls from using past results.  Please see `test-cases/final-dump/need.wybe` for an example of how to use this.  You'll need calls to the benchmarking proc to prefix the code to be timed with `{impure}`.

I suggest renaming the `measure` proc to something like `cpu_time`, because we may want procs to get elapsed time or system time, as well as ones that produce statistics about GCs.  It would be convenient to also provide a `benchmark` proc that takes a closure, runs it, and prints out the total cpu time used.

File: test-cases/execution/benchmark_example.wybe (hunk 3 above)
Comment (pschachte) [reply test-cases/execution/benchmark_example.wybe:24]: Now (in the latest merge into master) marking a proc as impure or semipure will stop unneeded calls in the body from being optimised away, or repeated calls from using past results. Please see test-cases/final-dump/need.wybe for an example of how to use this. You'll need calls to the benchmarking proc to prefix the code to be timed with {impure}.

This unfortunately won't fix the problem that outputs produced by a closure can't be relied upon to have been assigned.

I suggest renaming the measure proc to something like cpu_time, because we may want procs to get elapsed time or system time, as well as ones that produce statistics about GCs. It would be convenient to also provide a benchmark proc that takes a closure, runs it, and prints out the total cpu time used.

**Comment by neon64** [comment]
> Now (in the latest merge into master) marking a proc as impure or semipure will stop unneeded calls in the body from being optimised away, or repeated calls from using past results. Please see test-cases/final-dump/need.wybe for an example of how to use this. You'll need calls to the benchmarking proc to prefix the code to be timed with {impure}.

Thanks for the heads-up!

I think there's something fishy going on with `{impure}` and higher-order functions though. I tried `{impure}` on the `cpu_time()` proc as well as `{impure}` on the higher-order function it calls, and I get this when compiling [benchmark_lib.wybe](https://github.com/pschachte/wybe/blob/a07b2e3c2d8619a1ece2d99f1eeb8073c9fbb377/test-cases/final-dump/benchmark_lib.exp#L202-L209).

```llvm
define fastcc void @"benchmark_lib.<0>"() section ".text.benchmark_lib.<0>" {
entry:
  %0 = tail call double @get_clock_time()
  %1 = tail call fastcc i64 @"benchmark_lib.#anon#1<0>"()
  %2 = tail call double @get_clock_time()
  %3 = tail call fastcc i64 @"benchmark.benchmark<0>"(i64 ptrtoint ([1 x i64]* @benchmark_lib.0 to i64))
  ret void
}
```

I was surprised to see that `cpu_time()` had been inlined even though it was marked as `{impure}`. Have I understood https://github.com/pschachte/wybe/pull/333 correctly - with the current implementation impure procs will be inlined under some circumstances? Unfortunately after LLVM optimisation this becomes:

```llvm
define fastcc void @"benchmark_lib.<0>"() local_unnamed_addr section ".text.benchmark_lib.<0>" {
entry:
  %0 = tail call double @get_clock_time()
  %1 = tail call double @get_clock_time()
  %2 = tail call fastcc i64 @"benchmark.benchmark<0>"(i64 ptrtoint ([1 x i64]* @benchmark_lib.0 to i64))
  ret void
}
```

which is obviously not correct, now there is no code in between the two timing points.

If I switch `cpu_time()` to be marked with `{impure, noinline}` then the example works as desired - since `cpu_time()` is no-longer inlined, it is defined in a separate LLVM module (`benchmark.wybe`), so LLVM isn't able to determine the result isn't used and optimize out the higher-order call. But this seems somewhat fishy.

pschachte:
I reversed that decision, so now impure procs are inlined by heuristic just the same as pure ones.  But the purity or otherwise of Wybe calls in the body of the impure proc is remembered after inlining, and still used to disable troublesome optimisations.  However, that only affects calls to Wybe procs; foreign calls in the body of a proc are not affected by marking the proc impure.  So you might need to mark some foreign calls impure, such as calls to C functions.  You can do that by putting `{impure}` before the foreign function name.  You should do that for any call to stateful C code, unless you thread the `io` state through.

pschachte (commented):
Hi Chris,

Is this ready to be merged?

PR #305, opened 2022-05-20
add --no-tcmc switch

neon64:
Just like there is a `--no-multi-specz` switch which disables multiple specialization, I have added a `--no-tcmc`  switch which disables the tail-call-modulo-construction optimization.

I realise this may end up being a bit of a rabbit-hole if we keep adding switches for every possible optimization the Wybe compiler might perform, but I guess it is quite similar to how e.g.: clang has a bunch of `-f` flags to control each possible optimization (https://clang.llvm.org/docs/ClangCommandLineReference.html#target-independent-compilation-options). Perhaps there is a different naming convention all of these flags should use? For now I have stuck with `--no-xyz`.

There was one trick with the implementation: we want to keep calling `fixupCallsInProc` even if the optimization is disabled, since we may need to "fixup" calls to external libraries (i.e.: the Wybe standard library), which did already get TCMC-optimized.

### DIFF
--- a/src/LastCallAnalysis.hs
+++ b/src/LastCallAnalysis.hs
@@ -12,7 +12,7 @@ module LastCallAnalysis (lastCallAnalyseMod, lastCallAnalyseProc) where
 import AST
 import qualified Data.List as List
 import qualified UnivSet
-import Options (LogSelection(LastCallAnalysis))
+import Options (LogSelection(LastCallAnalysis), Options (optNoTCMC))
 import Util (sccElts, lift2)
 import Data.Foldable (foldrM)
 import Data.List.Predicate (allUnique)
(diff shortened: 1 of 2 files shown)


jimbxb:
Perhaps instead of adding more and more boolean flags, we can maintain a set of optimisation flags, which themselves are an enumeration of the optimisation flags we want to support. 

At least that way we don't bloat the poor `Options` type more

pschachte (approved):
James's suggestion is a good one.  Except rather than naming all the optimisation options "No"Something, they should just be named after the feature, and there should be a default set of optimisations.  Then the option parser should understand that options that begin with "no-" mean to turn off the optimisation with the "no-" stripped off.  That would let us add some optimisations that are not enabled by default.

But I'm going to merge this now anyway.  I imagine this will make Chris's benchmarking easier.  We can clean up the optimisation options later.

@neon64 Can you please add an issue requesting this rationalisation of the optimisation options?  If you get a chance to implement it, too, a PR would be most welcome :-).

PR #313, opened 2022-05-23
Refactored recognition of TestBools

### LINKED ISSUE #307: Calling an impure higher-order proc without `!` generates an internal error
Here's an example test case which triggers the error:

```wybe
pub def measure(func: {impure}(), ?seconds_elapsed: float) {
    func()
    ?seconds_elapsed = 3.0
}

?l = [1,2,3]
measure({impure} { ?l2 = l }, ?time)
```

### Expected Output

A nice error message similar the following (which occurs when you try to call a non-higher-order proc without `!`

```
testing/higher_order_call_impure.wybe:12:1: Calling impure proc higher_order_call_impure.someimpurefunction<0> without ! non-purity marker
```

### Actual Output

This is part of the output with `-l Flatten`, showing the AST immediately before typechecking. You can see there is `testbool func` instead of `!func`.

```
AFTER FLATTENING:
 Module higher_order_call_impure
  representation  : (not a type)
  public submods  :
  public resources:
  public procs    : higher_order_call_impure.<0>
                    higher_order_call_impure.measure<0>
  imports         : use wybe
  resources       :
  procs           :

module top-level code > public {impure} (0 calls)
0: () use !wybe.io.io:
    [](?tmp#3 @higher_order_call_impure:6:12)
    [|](3 @higher_order_call_impure:6:11, tmp#3, ?tmp#2 @higher_order_call_impure:6:10)
    [|](2 @higher_order_call_impure:6:9, tmp#2, ?tmp#1 @higher_order_call_impure:6:8)
    [|](1 @higher_order_call_impure:6:7, tmp#1, ?tmp#0 @higher_order_call_impure:6:6)
    =(?l @higher_order_call_impure:6:2, tmp#0)
    measure({impure}{=(?l2 @higher_order_call_impure:7:21, l @higher_order_call_impure:7:26)} @higher_order_call_impure:7:9, ?time @higher_order_call_impure:7:32)


measure > public (0 calls)
0: measure(func:{impure}(), ?seconds_elapsed:float):
    testbool func
    =(?seconds_elapsed @higher_order_call_impure:3:6, 3.0 @higher_order_call_impure:3:24)

LLVM code       : None

Internal error: testbool func:wybe.bool in a Det context
  error, called at src/AST.hs:4089:17 in main:AST
```

jimbxb:
Closes #307 

TestBools have been removed from Flatten, as there is syntactic ambiguity with some higher order calls. 
This is now checked in Types, as the type of the variable can be known.

Relevant test cases have been added

### DIFF
--- a/test-cases/final-dump/higher_order_impure_fail.exp
+++ b/test-cases/final-dump/higher_order_impure_fail.exp
@@ -1,0 +1,4 @@
+Error detected during type checking of module(s) higher_order_impure_fail
+[91mfinal-dump/higher_order_impure_fail.wybe:2:5: Calling impure higher-order term func, expecting at least semipure
+[0m[91mfinal-dump/higher_order_impure_fail.wybe:2:5: Calling impure higher-order term func without ! non-purity marker
+[0m
--- a/test-cases/final-dump/purity_warning.exp
+++ b/test-cases/final-dump/purity_warning.exp
@@ -33,5 +33,5 @@ define external fastcc  void @"purity_warning.<0>"()    {
 entry:
   ret void 
 }
-[93mfinal-dump/purity_warning.wybe:1:2: Calling proc wybe.list.length<0> with unneeded ! marker
+[93mfinal-dump/purity_warning.wybe:1:2: Calling proc proc wybe.list.length<0> with unneeded ! marker
 [0m
--- a/src/Flatten.hs
+++ b/src/Flatten.hs
@@ -50,8 +50,6 @@ import Control.Monad
 import Control.Monad.Trans.Class
 import Control.Monad.Trans.State
 import Control.Monad.Trans (lift,liftIO)
-import Distribution.Simple.Test.Log (TestSuiteLog(logFile))
-import Data.Type.Equality (inner)
 
 
 ----------------------------------------------------------------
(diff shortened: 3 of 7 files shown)


test-cases/final-dump/purity_warning.exp
```diff
@@ -33,5 +33,5 @@ define external fastcc  void @"purity_warning.<0>"()    {
 entry:
   ret void 
 }
-[93mfinal-dump/purity_warning.wybe:1:2: Calling proc wybe.list.length<0> with unneeded ! marker
+[93mfinal-dump/purity_warning.wybe:1:2: Calling proc proc wybe.list.length<0> with unneeded ! marker
```
neon64, line 36: is the "proc proc" a typo?

test-cases/final-dump/higher_order_impure_fail.exp
```diff
@@ -0,0 +1,4 @@
+Error detected during type checking of module(s) higher_order_impure_fail
+[91mfinal-dump/higher_order_impure_fail.wybe:2:5: Calling impure higher-order term func, expecting at least semipure
```
neon64, line 2: Hey I just noticed there's this 2nd error:
`final-dump/higher_order_impure_fail.wybe:2:5: Calling impure higher-order term func, expecting at least semipure`

Was it intentional to leave off the `{impure}` marker from [higher_order_impure_fail.wybe](test-cases/final-dump/higher_order_impure_fail.wybe), vs. having it in [higher_order_impure.wybe](test-cases/final-dump/higher_order_impure.wybe)? This test case is testing two separate errors in the code now I guess, rather than just `Calling impure higher-order term func without ! non-purity marker`

neon64 (commented):
Hey! Thanks heaps for working on a fix for this one! (plus all the other little papercuts I found over the last few days)

test-cases/final-dump/purity_warning.exp
jimbxb: Oops.

test-cases/final-dump/higher_order_impure_fail.exp
jimbxb: Yeah, that was intentional. The original code that raised the error actually had these two issues, so I wanted to catch both of them

pschachte (approved):

PR #315, opened 2022-05-24
Remove redundant getelementptr from gcAccess

neon64:
Putting this PR up as a draft so I don't forget. I will try to get it merged after I've finished writing my report.

Consider the following excerpt of LLVM IR:
```
  %5 = add   i64 %2, 8 
  %6 = inttoptr i64 %5 to i64* 
  %7 = getelementptr  i64, i64* %6, i64 0 
  store  i64 0, i64* %7 
```

I believe `getelementptr` here is redundant. We are offset a pointer `%6`, selecting the 0-th field, which returns the exact same pointer as we just passed in.

### DIFF
--- a/src/Blocks.hs
+++ b/src/Blocks.hs
@@ -447,10 +447,10 @@ codegenBody body = do
             M.void $ ret =<< buildOutputOp
         PrimFork var ty _ fbody -> do
             cgen prims False
-            codegenFork var ty fbody 
+            codegenFork var ty fbody
 
 
--- | Code generation for a conditional branch. 
+-- | Code generation for a conditional branch.
 -- XXX revise when LPVM can be transformed to have n>3-ary branches
 codegenFork :: PrimVarName -> TypeSpec -> [ProcBody] -> Codegen ()
 codegenFork _ _ [] = shouldnt "fork with no branches"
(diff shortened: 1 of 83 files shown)


jimbxb:
Bumping this.

I believe the LLVM optimiser removes this instruction, but there's no need for use to generate it in the first place. Good spot

neon64:
> I believe the LLVM optimiser removes this instruction

yep it sure does! This really just makes it easier to read for whenever I have to manually inspect the `.ll` output.

I've taken this PR out of draft now

jimbxb:
I am cautious though. Isn't it more idiomatic to use a GEP instead of adding to pointers?

pschachte (approved):
I'm happy to merge this as long as it doesn't break the generated binary code or upset the LLVM compiler.  Generating simpler LLVM code is a clear improvement, and if we ever decide to let LLVM lay out structures for us, we can go back to using GEP.

One other advantage of the Wybe structure layout code is that it orders the fields to pack them better.  Eg, if you create a structure that interleaves chars with ints, the Wybe compiler will pack the chars together, while I believe the LLVM compiler will keep them in the declared order, filling the structure with holes.  It's hard to see how to share the work of structure layout, other than for to give all the responsibility to either the Wybe compiler or LLVM.

PR #326, opened 2022-06-03
Add LPVM `void` instruction

### LINKED ISSUE #325: Add an LPVM construct to subvert some pure procedures from being removed.
This came up in an older rendition of global variables wherein I had an LPVM `void` instruction that existed solely to avoid a pure call from being omitted, assuming it is marked as being `semipure`/`impure`.

I.e., something like the following

```
complex_computation(?x)
foreign lpvm {impure} void(x)
```

This would force the compiler to produce a program that does said `complex_computation` without having to make the `complex_computation` impure. We shouldn't explicitly generate the `void` instruction when translating to LPVM, as its just a dummy (perhaps the right name to use is `phantom` :-) ).

I could see this being useful in the benchmarking module (#304), as you may want to benchmark a procedure that returns something but don't care about the returned value. The LLVM compiler might have something to say about removing the call, but I suspect it will be okay

jimbxb:
Closes #325

### DIFF
--- a/src/Blocks.hs
+++ b/src/Blocks.hs
@@ -842,6 +842,12 @@ cgenLPVM "load" _ args = do
         _ ->
             shouldnt $ "lpvm load instruction with wrong arity " ++ show args
 
+cgenLPVM "void" _ args = 
+    case partitionArgs args of
+        (_, outs@(_:_)) -> do
+            shouldnt $ "lpvm void cannot produce outputs " ++ show outs
+        _ -> return ()
+
 cgenLPVM pname flags args = do
     shouldnt $ "Instruction " ++ pname ++ " arity " ++ show (length args)
                ++ " not implemented."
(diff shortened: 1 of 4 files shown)


pschachte (commented):
Can't this be done more simply with an semipure library proc like:
```
pub def {semipure} need(x:T) { pass }
```
That ought to ensure  its argument is treated as needed without generating any code itself.

(I don't think `void` is a good name, as it does't convey its purpose.)

jimbxb:
I think `need` would have to be marked as `noinline` also, but it could work.

I chose the name `void` as the value is "cast into the void". The instruction also has the benefit of being variadic in the number of arguments, which isn't terribly useful, but I am sure its at least useful somewhere

pschachte:
Actually, no, it doesn't work, even with noinline, because need is (ironically) inferred not to need its argument.  Because it's semipure, it's not removed, but its argument is still considered to be unneeded, so it doesn't do what is needed.

jimbxb:
Ironic. 

Perhaps we can call the instruction `need` then?

pschachte:
I think that would be a better name.  But I'd prefer not to introduce a new instruction for that purpose; I'd rather find a way to use impurity, on the theory that it's better to put the solution in the library than the abstract machine.

If I can't work out a way to do that, we can add a new instruction.

jimbxb:
Sure. Closing for now then.

I'm unsure we can come up with something, however. It seems like a very special case, as the whole idea behind the abstract machine is to get rid of anything that isn't needed. Hence, having this dummy instructions solves the problem.

PR #328, opened 2022-06-05
Added LLVM FunctionAttributes to defined functions

### LINKED ISSUE #327: Translate `noinline` and `inline` modifiers into LLVM
I noticed when writing the code for #325 the LLVM compiler would inline procedures that are marked as `noinline`. LPVM follows these modifies, though the LLVM compiler does not have this information provided to it. 

More information on the LLVM compiler's function attributes (that control inlining) can be found here https://llvm.org/docs/LangRef.html#function-attributes

jimbxb:
Closes #327 

Note: May be more appropriate to use the `InlineHint` attribute over the `AlwaysInline` hint. `AlwaysInline` is a much stronger attribute, and may be an issue for recursive procs

### DIFF
--- a/src/Blocks.hs
+++ b/src/Blocks.hs
@@ -48,6 +48,7 @@ import qualified LLVM.AST.IntegerPredicate       as IP
 import           LLVM.AST.Operand                hiding (PointerType, operands)
 import           LLVM.AST.Type
 import           LLVM.AST.Typed
+import qualified LLVM.AST.Attribute              as A (FunctionAttribute(..)) 
 import           LLVM.Pretty                     (ppllvm)
 
 import qualified Data.ByteString                 as BS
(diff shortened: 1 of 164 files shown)


pschachte (approved):
A LPVM proc that's marked for inlining should never be called (since all its calls will be inlined).  At link time, the generated machine code for them should be stripped out.  Now we could leave inlining out of the LPVM code and just let the LLVM compiler do the inlining, but inlining at the LPVM level gives more opportunities for optimisation, so I don't think that's a good idea.  As it is, the LPVM code for inlined procs is only needed to allow calls to them to be inlined, but the LLVM code for them should not be needed.  I think a better fix for inlined procs would be simply not to generate LLVM code for them at all.

The point about noinline is quite right, though:  if you don't want them to be inlined at the LPVM level, then you don't want them to be inlined by the LLVM compiler, either.  So that part of this change is right, and the change to mark LLVM procs for inlining is harmless, so I'll merge this PR anyway.

PR #333, opened 2022-06-17
 Don't automatically inline impure procs.

### LINKED ISSUE #332: Impure procs should not be inlined unless explicitly marked for inlining
... because inlining loses impurity information.

For example

```
def {impure} impure {foreign c {impure} printf(c"Hello, World!\n")}

!impure # shouldn't be inlined
```

pschachte:
Fix #332


jimbxb:
I think we can do better than this. It shouldn't hurt to inline a proc if the purity of the callee is no greater than the caller. I.e. inlining a pure proc in a semi pure context is okay, as is inlining an impure proc in an impure context. 

So perhaps we could add another level of inlining, where if inlining is inferred we can only inline if the purity of the callee is no greater.

jimbxb:
I think it might be necessary to attach the purity information to HO calls, also. While HO calls do have purity information through the HO term's HO type modifiers, this is equivalent information to the purity of the callee we currently have for FO calls, but distinct from this new purity information we have from the caller. 

I think that makes sense. Some HO calls may be made in a non-pure context but have no needed outputs, and I think we'd not want those calls to be omitted as is the case for FO calls from inlined procs.

jimbxb:
Another thing to check is to make sure the optimised LLVM is behaving as we expect. `--dump-opt-llvm` on that new test case should show us what we need to know.

jimbxb:
Actually, now that I look at it, it seems that the top-level call to need has been omitted. Was this intended?

pschachte:
Good point about both issues.

I've added impurity info to HO calls, and that actually fixed the problem with the `need` test case, since now the HO call to `stmt` has the maximum impurity of `need` (semipure) and `stmt` (impure), and even after `need` is inlined, it retains that impurity.

With this definition of `need`, you can put `!need {impure}` in front of a chunk of code in braces to make sure the code doesn't get optimised away.  And you can use that in a proc of any level of purity.

src/Blocks.hs
```diff
@@ -546,20 +546,21 @@ cgen (prim@(PrimCall callSiteID pspec args _):nextPrims) isLeaf = do
...
     cgen nextPrims' isLeaf
 
-cgen (prim@(PrimHigher cId (ArgClosure pspec closed _) args):ps) isLeaf = do
+cgen (prim@(PrimHigher cId (ArgClosure pspec closed _) _ args):ps) isLeaf = do
     pspec' <- fromMaybe pspec <$> lift2 (maybeGetClosureOf pspec)
     logCodegen $ "Compiling " ++ show prim
               ++ " as first order call to " ++ show pspec'
               ++ " closed over " ++ show closed
-    cgen (PrimCall cId pspec' (closed ++ args) univGlobalFlows : ps) isLeaf
+    cgen (PrimCall cId pspec' Pure (closed ++ args) univGlobalFlows : ps) isLeaf
```
jimbxb, line 554: This shouldn't be `Pure`. I suppose `univGlobalFlows` is also incorrect too.

src/Clause.hs
```diff
@@ -179,48 +180,48 @@ compileProc proc procID =
 --  statement or, in the case of a SemiDet proc, as the final statement of a
 --  body. This code assumes that these invariants are observed, and does not
 --  worry whether the proc is Det or SemiDet.
-compileBody :: [Placed Stmt] -> [Param] -> Determinism -> ClauseComp ProcBody
-compileBody [] params detism = do
+compileBody :: Impurity -> [Placed Stmt] -> [Param] -> Determinism -> ClauseComp ProcBody
```
jimbxb, line 183: Can we pass Impurity as a field in the state as opposed to a parameter?

jimbxb (changes requested):

jimbxb:
I am also curious, now that prims have their context's purity attached, is there a reason for disallowing `impure` procs to be automatically inlined?

src/Clause.hs
pschachte: Good idea.  Done.

src/Blocks.hs
pschachte: It doesn't really matter, does it?  `cgen` ignores thsose two arguments in a `PrimCall`.  Once we get to generating LLVM, we stop worrying about purity.

jimbxb: Yeah, it doesn't. Though my concern is that if in the future there's an extra step that needs to know about the impurity this could be easily missed. It's a small change but could be a timesaver in the future

pschachte: Fair point.  I'll fix it.  I'm not sure how to fix `univGlobalFlows`, though.  I guess it could be the universal set or the empty set.  Can you fix that?

jimbxb: That can be found using `getProcGlobalFlows`

pschachte:
> I am also curious, now that prims have their context's purity attached, is there a reason for disallowing `impure` procs to be automatically inlined?

I'm not sure what's the right thing to do here.  My argument was that when you say a proc is impure, you're saying "whoa, be careful optimising this", and that would include being careful about inlining it.  If we allow autoinlining of impure code, your test case

```
pub def {impure} measure(func: {impure}(), ?seconds_elapsed: float) {
    !func()
    ?seconds_elapsed = 3.0
}

?l2 = [1,2,3]
!measure({impure} { ?l = l2 }, ?time)
!println(time)
```
generates this output for the closure:
```
gen#1 > {inline,impure} (1 calls)
0: higher_order_impure.gen#1<0>
gen#1([^l2##0:wybe.list(wybe.int)])<{}; {}>:
  AliasPairs: []
  InterestingCallProperties: []
gen#1 > {inline,impure} (1 calls)
1: higher_order_impure.gen#1<1>
gen#1([^l2##0:wybe.list(wybe.int)])<{}; {}>:
  AliasPairs: []
  InterestingCallProperties: []
```
and this for main:
```
module top-level code > public {inline,semipure} (0 calls)
0: higher_order_impure.<0>
()<{<<wybe.io.io>>}; {<<wybe.io.io>>}>:
  AliasPairs: []
  InterestingCallProperties: []
    higher_order_impure.gen#1<0>(_:wybe.list(wybe.int)) #0 @higher_order_impure:nn:nn
    foreign lpvm load(<<wybe.io.io>>:wybe.phantom, ?%tmp#18##0:wybe.phantom) @float:nn:nn
    foreign c print_float(3.0:wybe.float, ~tmp#18##0:wybe.phantom, ?tmp#19##0:wybe.phantom) @float:nn:nn
    foreign c putchar('\n':wybe.char, ~tmp#19##0:wybe.phantom, ?tmp#20##0:wybe.phantom) @io:nn:nn
    foreign lpvm store(~%tmp#20##0:wybe.phantom, <<wybe.io.io>>:wybe.phantom) @io:nn:nn
```
so the call to `!measure({impure} { ?l = l2 }, ?time)` doesn't do anything for the closure.  That's because the `=` call turns into an LLVM move, and LLVM instructions are not treated as impure and are optimised as usual.  Do you want to consider this a bug in the test case, because the closure is too simple?  Or is this the output you would expect?  Or is this a problem with inlining impure procs?

jimbxb:
I think I'd expect that. The higher order term is already optimised as that currently, and I would probably expect that to be the case as it's never needed, yet it might be confusing as the `=` doesn't look like a foreign call.

Btw, did you intend to delete the `higher_order_impure` test case?

WYBE.md
```diff
@@ -1715,6 +1717,8 @@ force inlining of calls to this proc
 - `noinline`
 prevent inlining of calls to this proc
 
+Procs that are declared `impure` are never automatically inlined, however, you may explicitly declare them `inline`.
```
jimbxb, line 1720: This has now changed, I believe.

jimbxb (approved):
Other than that, looks fine by me. :)

WYBE.md
pschachte: Yup, good catch.

src/BodyBuilder.hs
```diff
@@ -403,7 +403,7 @@ instr' prim@(PrimForeign "llvm" "move" []
 --     the only problem that needs fixing.  We don't want to fix this
 --     by threading a heap through, because it's fine to reorder calls
 --     to alloc.
--- XXX can we git rid of this special case by making lpvm alloc "impure"?
+-- XXX can we git rid of this special case by making lpvm alloc "semipure"?
```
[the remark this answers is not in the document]
jimbxb: Actually now that I think about it, that doesn't seems so bad. If you ask for memory and then you use it twice, chances are one of them won't do so destructively and youll just have to do two mallocs anyway... But that would also implicitly call `memcpy`. *Sigh*

PR #339, opened 2022-06-25
Proof-of-concept: Inferred inline

jimbxb:
See discussion in #332

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -53,7 +53,8 @@ module AST (
   emptyInterface, emptyImplementation,
   getParams, getPrimParams, getDetism, getProcDef, getProcPrimProto,
   mkTempName, updateProcDef, updateProcDefM,
-  ModSpec, maybeModPrefix, ProcImpln(..), ProcDef(..), procInline, procCallCount,
+  ModSpec, maybeModPrefix, ProcImpln(..), ProcDef(..), 
+  procInline, procUserInline, procInferredInline, procCallCount,
   getProcGlobalFlows,
   primImpurity, flagsImpurity, flagsDetism,
   AliasMap, aliasMapToAliasPairs, ParameterID, parameterIDToVarName,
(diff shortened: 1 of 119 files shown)


src/AST.hs
```diff
@@ -1096,15 +1097,23 @@ data ProcModifiers = ProcModifiers {
 } deriving (Eq, Ord, Generic)
 
 
-data Inlining = Inline | MayInline | NoInline
+data Inlining = Inline
```
pschachte, line 1100: OK, so the idea here is that `MaybeInline` means the user hasn't specified any preference about inlining and `InferredInline` means the same, except that the compiler has decided that inlining would be worthwhile.  That's nice and declarative, allowing the compiler to decide to ignore the `InferredInline` and still not inline if the proc is impure.

However, I've been planning to have the LLVM code generator not generate LLVM code for inlined procs, on the theory that they should never be called at runtime.  If we use characteristics of the caller to decide whether to inline, we couldn't omit code generation for inline procs.  That wouldn't be a huge loss, but it's a loss.  Also, it would make the compiler's treatment of impure code less predictable for the programmer.  So the question is:  what is gained by taking the caller's purity into account when deciding whether to inline a call?

jimbxb: I'm not sure we can omit all inlined procs in LLVM because of the case of recursive procs. If a proc calls itself, or even if a proc calls another that calls the original proc, no matter how we inline, the inlining may keep calls to an inlined procs. Hence, we can't omit *all* of them in LLVM.

I don't think it would make the behaviour less predictable if we document it. If you're using impure code already, I'd think that what the compiler does with the code is already fairly unpredictable because impure code is still subject to the same optimisations that pure code is, sans omission.

The one benefit we do gain with this `InferredInline` is that, unless the programmer asks for the proc to be inlined, we can be sure that the purity information is never lost, which was the primary issue with #332.

pschachte: I'm not sure inlining of recursive procs makes sense, because it increases code size and I'm not sure it actually speeds anything up, since it will have to execute a conditional branch and in all likelihood a call anyway.  What you'd actually want to do for a small recursive proc would be to turn it into an explicit loop at the LLVM level and then inline that, turning the call into an inline loop.  Inlining a recursive proc at the LPVM level won't do that, and will actually make that harder to do.  I'm not sure, but it might work if it's tail recursive and marked for inlining at the LLVM level.

So I was assuming we wouldn't have any recursive procs marked for inlining at the LPVM level.  Or, if the LLVM inlining of tail recursive procs works, we could create a new inlining state called `InlineRecursive` which would not be inlined at the LPVM level, but would trigger the generated LLVM code to be marked for inlining.  It'd be cool for procs like map and filter (and many more) to be inlined as loops.

As for predictability, I'm working on disabling at least some optimisations inside impure procs.  I'll keep optimisations of LLVM instructions, but disable most or all optimisations of pure instruction inside impure procs.  That gives not only predictability, but also control.

The approach I'm working on to fix the issue raised #332 even after inlining is to attach an `Impurity` to every `PrimCall`.  The impurity will be the greater of the impurity of the proc being called and the proc the call appears in.  That impurity is used to determine what optimisations to perform on the call.

PR #344, opened 2022-07-05
Unified resource transformations

jimbxb:
This PR unifies the resource transformation passes that transform resources into params/args and then globals. Uniqueness checking interfered with this before, however, now I have changed that pass to handle pre-resource transformation code by using the resources found in a proc's prototype instead of the parameterised resources.

This unification isn't strictly necessary, however it does simplify the resource module greatly.

### DIFF
--- a/src/Resources.hs
+++ b/src/Resources.hs
@@ -9,7 +9,7 @@
 
 module Resources (resourceCheckMod, canonicaliseProcResources,
                   canonicaliseResourceSpec,
-                  transformProcResources, transformProcGlobals) where
+                  transformProcResources) where
 
 import           AST
 import           Control.Monad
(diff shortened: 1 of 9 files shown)


jimbxb:
Converting to draft to make a few little tweaks

src/Unique.hs
```diff
@@ -82,27 +83,38 @@ initUniquenessState = UniquenessState Map.empty [] False
 -- | Check correctness of uniqueness for a procedure
 uniquenessCheckProc :: ProcDef -> Int -> Compiler ProcDef
 uniquenessCheckProc def _ = do
-    let name = procName def
+    let name = showProcName $ procName def
     let pos = procPos def
-    logMsg Uniqueness $ "Uniqueness checking proc: " ++ showProcName name
+    logMsg Uniqueness $ "Uniqueness checking proc: " ++ name
```
pschachte, line 88: This change means that module top level code just has the empty string after "Uniqueness checking proc:".  Keeping the call to `showProcName` would give more intelligible logging.

src/Resources.hs
```diff
@@ -454,8 +243,7 @@ globaliseProc pos name variant detism params ress body = do
     -- Unbranching will then force the stores to be executed in the case of
     -- failure, ensuring the global variables are unmoddified
     newMentioned <- gets resMentioned
-    body'' <- if detism /= SemiDet && detism /= Failure
-                || Map.null newMentioned
+    body'' <- if detism /= SemiDet && detism /= Failure || Map.null newMentioned
```
pschachte, line 246: It would be better to define a function `determinismCanFail :: Determinism -> Bool` and use that.  I expect we will eventually introduce other determinisms that can fail.

pschachte (approved):
A couple of minor issue, then it's fine to merge.

PR #345, opened 2022-07-05
Reduced continuation inputs; modified gen-sym'd proc names

jimbxb:
This PR removes definitely unneeded parameters from continuation procs, as mentioned in #87 

I also took the liberty of changing names that are generated for continuations to make it a little easier to know where they came from. The format is now ***proc***`#cont#`***N*** where ***proc*** is the name of the current proc and ***N*** is a monotonic number for the given combination of ***proc***`#cont`. Anonymous procedures and closures have also been given nicer names, being `anon` and `closure`, respectively, with similar increasing numbers

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -69,7 +69,7 @@ module AST (
   flowsIn, flowsOut, primFlowToFlowDir, isInputFlow, isOutputFlow,
   foldStmts, foldExps, foldBodyPrims, foldBodyDistrib,
   expToStmt, seqToStmt, stmtsImpurity, stmtImpurity, procCallToExp,
-  expOutputs, pexpListOutputs, expInputs, pexpListInputs,
+  stmtsInputs, expOutputs, pexpListOutputs, expInputs, pexpListInputs,
   setExpTypeFlow, setPExpTypeFlow,
   Prim(..), primArgs, replacePrimArgs, argIsVar, argIsConst, argIntegerValue,
   varsInPrims, varsInPrim, varsInPrimArgs, varsInPrimArg,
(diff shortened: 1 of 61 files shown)


src/Unbranch.hs
```diff
@@ -679,7 +681,7 @@ unbranchExp exp@(Closure ps free) pos = do
     isClosure <- lift $ isClosureProc ps
     if isClosure
     then return $ maybePlace exp pos
-    else newProcName >>= addClosure ps free pos
+    else newProcName "closure" >>= addClosure ps free pos
```
pschachte, line 684: Can this cause a problem if someone names a proc "closure" (or any other legal name)?

src/AST.hs
```diff
@@ -782,11 +782,12 @@ getModuleImplementationMaybe fn = do
...
-genProcName = do
-  ctr <- getModule procCount
-  updateModule (\mod -> mod {procCount = ctr + 1 })
-  return $ specialName2 "gen" $ show (ctr + 1)
+genProcName :: ProcName -> Compiler ProcName
+genProcName pname = do
+  names <- getModule procNames
+  let ctr = 1 + Map.findWithDefault 0 pname names 
+  updateModule (\mod -> mod {procNames = Map.alter (const $ Just ctr) pname names })
+  return $ specialName2 pname $ show ctr
```
pschachte, line 790: Are there potential collision problems with procs having multiple continuations, and the continuations having their own continuations, and so on?

jimbxb: This `ctr` number is appended to the proc's name each time a new name is requested. This `ctr` is unique to each proc name and increments each time a name is requested, much like the current `gen` names, so there's no chance of a collision.

src/Unbranch.hs
jimbxb: No. The name for a closure of a proc `foo` will be `foo#closure#1`, etc., which cannot clash with source defined names

src/Unbranch.hs
```diff
@@ -659,7 +661,7 @@ unbranchExp (Typed exp ty cast) pos = do
 unbranchExp exp@(AnonProc mods params pstmts clsd res) pos = do
     let clsd' = trustFromJust "unbranch annon proc without closed" clsd
     let res' = trustFromJust "unbranch annon proc without resources" res
-    name <- newProcName
+    name <- newProcName "anon"
```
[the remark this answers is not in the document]
jimbxb: See comments below

PR #350, opened 2022-07-11
Allow calls to impure functions within expressions

neon64:
While I was trying to work on https://github.com/pschachte/wybe/pull/304, I discovered what I think is a bug in the parser? You can write `!function_name()` as part of an expression, but by the time it gets to typechecking the `!` is silently lost, and you get this error:

```
testing/impure_fib.wybe:9:14: Calling impure proc impure_fib.fib<0> without ! non-purity marker
```

which I was very confused by, since I *did* use the `!` non-purity marker.

This PR adds a "uses resources" boolean flag to the `Fncall` Exp variant, mimicking the "uses resources" boolean flag on the `ProcCall` Stmt variant. This means the `!` marker is no-longer silently lost before typechecking the expression.

I also found a couple more functions in AST.hs which weren't used, which I removed since they're not actively being tested. I can revert this though if you'd prefer.

### DIFF
--- a/src/Flatten.hs
+++ b/src/Flatten.hs
@@ -282,14 +282,14 @@ flattenStmt' stmt@(ProcCall fn@(First [] "=" id) callDetism res [arg1,arg2]) pos
     logFlatten $ "Flattening assignment with arg1 outputs? " ++ show arg1outs
                  ++ " and arg2 outputs? " ++ show arg2outs
     case (arg1content, arg2content) of
-      (Fncall mod name args, _)
+      (Fncall mod name resourceful args, _)
         | not arg1outs && arg2outs -> do
-        let stmt' = ProcCall (First mod name Nothing) Det False (args++[arg2])
+        let stmt' = ProcCall (First mod name Nothing) Det resourceful (args++[arg2])
         flattenStmt stmt' pos detism
-      (_, Fncall mod name args)
+      (_, Fncall mod name resourceful args)
         | not arg2outs && arg1outs -> do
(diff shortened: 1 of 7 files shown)


src/Flatten.hs
```diff
@@ -596,8 +596,8 @@ flattenExp (CaseExp pexpr cases deflt) ty castFrom pos = do
     pexpr' <- flattenPExp pexpr
     flattenStmt (translateExpCases pexpr' resultName cases deflt) pos Det
     return $ maybePlace (Var resultName ParamIn Ordinary) pos
-flattenExp (Fncall mod name exps) ty castFrom pos = do
-    let stmtBuilder = ProcCall (First mod name Nothing) Det False
+flattenExp (Fncall mod name resourceful exps) ty castFrom pos = do
+    let stmtBuilder = ProcCall (First mod name Nothing) Det resourceful
```
neon64, line 600: This is the crux of the change. Previously, all `Fncalls` had their "can use resources" flag hardcoded to `False`.

jimbxb:
I'm not sure we want to allow this. See #343 for basically identical changes that strictly disallows `!` on expression function calls. 

The reasoning behind this change is because it's not exactly which argument should be evaluated first, as is the case for a resourceful term. You can always use a `where` clause to use an impure function, too.

pschachte:
I agree with James, we shouldn't allow non-pure function calls, because they do not establish the ordering of effects.

neon64:
Ah, I see now how #343 is 90% the exact same PR, except for disallowing `!` in this case. Will close this one

PR #351, opened 2022-08-07
Added warning when type constraint is of the current type

### LINKED ISSUE #135: Issue a warning for explicitly using the current module name as a type, instead of `_`
This ensures that it's possible to rename the file a module is defined in without breaking the code.

jimbxb:
Fixes #135 

This was one of the first issues I attempted to tackle. It turns out the issue is quite deep-rooted. 

This also adds a lot of positioning to auto-generated procs, which is useful for these warnings. If not, these warnings that arise from (de)constructors/getters/mutators are unplaced.

### DIFF
--- a/test-cases/final-dump/generic_list.exp
+++ b/test-cases/final-dump/generic_list.exp
@@ -348,23 +348,13 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
 entry:
   ret i64 0 
 }
-[93mExplicit specification of current type generic_list(T),
+[93mfinal-dump/generic_list.wybe:1:40: Explicit specification of current type generic_list(T),
   it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:1:27: Explicit specification of current type generic_list(T),
-  it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:1:40: Explicit specification of current type generic_list(T),
-  it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:3:5: Explicit specification of current type generic_list(T),
+[0m[93mfinal-dump/generic_list.wybe:3:9: Explicit specification of current type generic_list(T),
   it is recommended to specify type as _
 [0m[93mfinal-dump/generic_list.wybe:3:16: Explicit specification of current type generic_list(T),
   it is recommended to specify type as _
 [0m[93mfinal-dump/generic_list.wybe:3:35: Explicit specification of current type generic_list(T),
-  it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:4:9: Explicit specification of current type generic_list(T),
-  it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:4:28: Explicit specification of current type generic_list(T),
-  it is recommended to specify type as _
-[0m[93mfinal-dump/generic_list.wybe:5:17: Explicit specification of current type generic_list(T),
   it is recommended to specify type as _
 [0m[93mfinal-dump/generic_list.wybe:8:16: Explicit specification of current type generic_list(T),
   it is recommended to specify type as _
--- a/test-cases/final-dump/generic_list.exp
+++ b/test-cases/final-dump/generic_list.exp
@@ -348,3 +348,26 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
 entry:
   ret i64 0 
 }
+[93mExplicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:1:27: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:1:40: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:5: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:16: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:35: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:4:9: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:4:28: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:5:17: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:8:16: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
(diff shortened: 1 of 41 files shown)


test-cases/final-dump/loop_bug.exp
```diff
@@ -490,3 +490,8 @@ entry:
   %1 = xor i1 %0, 1 
   ret i1 %1 
 }
+[93mfinal-dump/loop_bug.wybe:1:31: Explicit specification of current type int_list,
```
pschachte, line 493: This warning looks erroneous.  It points to the `cons` constructor.  The second warning points to the tail member, which should be warned about, but the constructor should not be.

test-cases/final-dump/generic_use.exp
```diff
@@ -871,3 +871,26 @@ if.then:
...
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:1:40: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:5: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:16: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:35: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:4:9: Explicit specification of current type generic_list(T),
```
pschachte, line 886: The `generic_list` type name doesn't appear on lines 4 or 5, so these warnings are spurious.

test-cases/final-dump/generic_list.exp
```diff
@@ -348,3 +348,26 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
 entry:
   ret i64 0 
 }
+[93mExplicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:1:27: Explicit specification of current type generic_list(T),
```
pschachte, line 353: Column 27 is the `cons` constructor, which doesn't explicitly mention the type name.

test-cases/final-dump/generic_list.exp
```diff
@@ -348,3 +348,26 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
...
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:1:40: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:5: Explicit specification of current type generic_list(T),
```
pschachte, line 357: Column 5 is the def keyword. Presumably this refers to the return type of append; in this case, it should probably reference column 9, which is the function name. Ideally the warning would explicitly mention that it refers to the function return type, if that's not too hard.

test-cases/final-dump/generic_list.exp
```diff
@@ -348,3 +348,26 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
...
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:16: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:3:35: Explicit specification of current type generic_list(T),
+  it is recommended to specify type as _
+[0m[93mfinal-dump/generic_list.wybe:4:9: Explicit specification of current type generic_list(T),
```
pschachte, line 363: Lines 4 and 5 don't mention the type name at all so there shouldn't be any warnings for those lines.

test-cases/final-dump/generic_list.exp
```diff
@@ -348,3 +348,26 @@ define external fastcc  i64 @"generic_list.nil<0>"() alwaysinline   {
 entry:
   ret i64 0 
 }
+[93mExplicit specification of current type generic_list(T),
```
pschachte, line 351: This warning doesn't have a line number, so I'm not sure what it refers to.

test-cases/final-dump/type_int.exp
```diff
@@ -274,3 +274,86 @@ entry:
...
 }
+[93mExplicit specification of current type myint,
+  it is recommended to specify type as _
+[0m[93mfinal-dump/type_int.wybe:2:9: Explicit specification of current type myint,
+  it is recommended to specify type as _
+[0m[93mfinal-dump/type_int.wybe:2:17: Explicit specification of current type myint,
+  it is recommended to specify type as _
+[0m[93mfinal-dump/type_int.wybe:2:25: Explicit specification of current type myint,
+  it is recommended to specify type as _
+[0m[93mfinal-dump/type_int.wybe:2:42: Explicit specification of current type myint,
```
pschachte, line 285: Column 42 is the `foreign` call, which doesn't explicitly specify the type anywhere.

pschachte (changes requested):
This is nice work, and has the pleasant side effect of giving better source positions in unrelated errors.  But there are a few spurious warning messages.

jimbxb:
Yep, the issues arise where I'm using the wrong source POS. I'll fix it up soon (I hope)

test-cases/final-dump/loop_bug.exp
jimbxb: The warning occurs because the field `tail` has been given an explicit type, and this type is propagated into the (de)constructor.

test-cases/final-dump/generic_use.exp
jimbxb: I believe these once come from the transformation of a `FuncDecl` into a `ProcDecl` with an explicitly typed `move`. I can get rid of those,

test-cases/final-dump/generic_list.exp
jimbxb: As per previous comment, the field does.

test-cases/final-dump/generic_list.exp
jimbxb: See `generic_use`.

test-cases/final-dump/generic_list.exp
jimbxb: Grumble. Will dig into this

jimbxb: Ah, there was an `Unplaced` in the aforementioned `move`!

test-cases/final-dump/type_int.exp
jimbxb: This looks to be fixed with the `FuncDecl` `move` typing.

test-cases/final-dump/loop_bug.exp
pschachte: If we just didn't issue this warning at all for generated accessor/mutator/deconstructor/equality procs/funcs (but do warn for generated constructor procs), would that eliminate any needed warnings?  In this case, you'd get a warning for the constructor, so the one for the deconstructor is redundant.  And you recently added information noting which procs are generated, so it should be possible.

test-cases/final-dump/generic_list.exp
jimbxb: I think this one requires a tweak of the parser. The position is grabbed from the `"def"` keyword, but I can get it from where the `"def"` token ends easily.

jimbxb: As for changing the message I think that'll be a bit more work, as I'd have to special case this one. I am not sure it is work it.

pschachte (approved):

PR #355, opened 2022-09-14
Cond consequences

pschachte:
Generate switches (computed gotos) for if-elseif-elseif...else conditionals that densely cover a set of integer values for a single variable over a range beginning with 0.


jimbxb:
Beautiful :)

pschachte:
:-)

Turned out to be much harder than expected, because of all the code between an else and the following if.

It still only works when all numbers between 0 and the maximum are covered.  I'll loosen that limitation in a later PR.

PR #356, opened 2022-09-27
Generate more intelligent code where fields are shared between constructors

### LINKED ISSUE #354: Allow a type to have identically named members in multiple constructors, and produce accessors and mutators that work for all appropriate constructors
In cases where *all* constructors have identically named members, the generated accessors/mutators should be det.

for example, a type

```
pub constructors book(title:string, author:string, genre:genre, id:int)
                           | magazine(title:string, frequency:publication_frequency, id:int)
                           | DVD(title:string, director:string, genre:genre, id:int)
```
should have det accessor and mutator procs named `title` and `id`, and semidet ones named `genre`.

jimbxb:
Closes #354


pschachte (approved):
Wonderful!  Thank you for that.

Could you also update the language doc to mention this?

PR #361, opened 2022-09-27
Updated docs for getters/setters; cleaned up the markdown file a little

jimbxb:


pschachte (changes requested):
Thanks for the cleanup.  Could you please add some text explicitly stating that multiple constructors for a single type may share a field name, in which case the accessor and mutator generated for that field will work to access or mutate that field for all the constructors that have a field with that name?

pschachte (approved):
Thanks.

PR #363, opened 2022-11-13
Add a return statement

jimbxb:
This is a work-in-progress attempt at adding a `return` statement. In essence, it is the det dual of `fail`, causing a call to exit, but the call succeeds.

Some things are left to consider/get working correctly:
- mode checking
  - currently return causes the determinism to be `terminal`, but I suspect this isn't correct. I want to say that nothing more can happen, but we haven't failed and `terminal` seemed the best fit here, but it isn't quite right, because there is more to go after the call
- use blocks
  - any used resources must be reset to their value before the `use` block if `return` is encountered

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -2422,6 +2422,7 @@ foldStmt' sfn efn val (Not negated) pos =
 foldStmt' sfn efn val (TestBool exp) pos = foldExp sfn efn val exp Nothing
 foldStmt' _   _   val Nop pos = val
 foldStmt' _   _   val Fail pos = val
+foldStmt' _   _   val Return pos = val
 foldStmt' sfn efn val (Loop body _ _) pos = foldStmts sfn efn val body
 foldStmt' sfn efn val (UseResources _ _ body) pos = foldStmts sfn efn val body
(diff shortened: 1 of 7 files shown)


pschachte (changes requested):
I don't think `terminal` is the right determinism for this.  Terminal means that it will neither succeed nor fail.  I think you'll need a new determinism that means something like "it has already succeeded so thoroughly that even if this is conjoined with other constraints, it automatically satisfies them".  But then you'll need to work out how this determinism conjoins, disjoins, and otherwise combines with, all the other determinisms.  This becomes tricky because none of the other determinisms is sensitive to procedure boundaries, and this one will need to be (procs shouldn't be declared to have this new determinism; and a proc all of whose branches have this determinism still has determinism `det`).

You can probably predict my next review comment:  WYBE.md needs to be revised to explain `return`, and to clearly explain how it interacts with everything else, including loops, but also what if a `return` appears in one of the conditions in an `if`.  There are probably other interactions I haven't thought of.  It would be good to have a motivating example for this feature.

I couldn't see how you're handling variable binding in case of a `return`.  It must be the case that all outputs of a proc are already bound when you reach a `return`, otherwise it must be reported as an error.

PR #365, opened 2022-11-17
Improved logging flags to toggle log aspects

jimbxb:
A small quality of life improvement. If you want to log all but say types, `-lno-types`.


pschachte (commented):
Nice.  That'll let you log everything but AST, for example.

PR #369, opened 2022-11-30
Added brainfuck parser/interpreter test-case

jimbxb:


pschachte (approved):
That's really nice!  I like the way you use type declarations, neatly modularising the code.

I'll merge this now.  But if you can be bothered, it would be great if you could add a bit of documentation to this and copy it to the `samples` directory I just created.

PR #371, opened 2022-12-30
Add foreign proc def

### LINKED ISSUE #366: Improve the foreign interface by allowing for foreign functions to be declared
Sometimes calling a foreign function can be a little clunky due to the repetitive nature of foreign calls. While this can be reduced by defining a Wybe "interface" as a series of functions that call their foreign counterpart, some language features could be used to resolve this. 

Currently one would have to do something like this:

```
def putchar(c:char) use !io {
    foreign c putchar(c, !io)
}

!putchar('W')
!putchar('y')
!putchar('b')
!putchar('e')

# or verbosely:
foreign c putchar('W', !io)
foreign c putchar('y', !io)
foreign c putchar('b', !io)
foreign c putchar('e', !io)
```

I see two potential solutions to this:

1. Declare a foreign function as you would a Wybe proc

This could look something like as follows:

```
# putchar can now be used as a "first-class" Wybe proc
def foreign c putchar(c:char) use !io

!putchar('W')
!putchar('y')
!putchar('b')
!putchar('e')
```

This really only reduces boilerplate for the foreign code, but I see it being useful. For resources, they should be passed as regular arguments to the foreign function, and probably in some canonical order -- if you want to change that order then you can just use the old school version as above.

2. Automatically create a module based on some foreign file. 

Peter has discussed this with me in the past, and I think his would be the best, but it certainly involves the most work (for good reason). In essence, given some foreign language file (perhaps an object file of some kind?) construct a Wybe module that acts as an interface to this module. This ideally would be performed in-source, to the effect of what follows:

```
foreign c module io

!putchar('W')
!putchar('y')
!putchar('b')
!putchar('e')
```

Of course, there would now have to be a C file that exists elsewhere that contains the relevant bindings, etc. for the module's procs, and I don't like the implicitness of it now that I'm writing it out. I'm not really sure how to approach this, which is why I believe the former is achievable.

jimbxb:
Closes #366

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -736,7 +736,7 @@ Wybe also supports an alternative syntax for invoking procedures, functions, or
 constructors that puts the (first) argument first, and the procedure, function,
 or constructor name, with its other arguments, if any, second:
 
-> *arg*`^`*operation*(*other args*)
+> *arg*`^`*operation*`(`*other args*`)`
 
 and in the special case of operations taking only one argument:
 
@@ -2064,7 +2064,7 @@ so be careful to get the call right.
 
 The form of a foreign call is:
 
-> `foreign` *language* *function*(*arg*, *arg*, ...)
+> `foreign` *language* *function*`(`*arg*, *arg*, ...`)`
 
 where *language* is the name of the foreign language the function is written in,
 *function* is the name of the foreign function to call, and the *arg*s are the
@@ -2092,7 +2092,7 @@ Foreign calls may optionally specify *modifiers* to provide extra information
 useful to the Wybe compiler.  If modifiers are to be specified, they are written 
 after the *language* name:
 
-> `foreign` *language* `{`*modifiers*`}` *function*(*arg*, *arg*, ...)
+> `foreign` *language* `{`*modifiers*`}` *function*`(`*arg*, *arg*, ...`)`
 
 where *modifiers* is a comma-separate sequence of identifiers specifying this
 information.  Supported modifiers in foreign calls are:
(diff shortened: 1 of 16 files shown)


jimbxb:
This is unfinished. I'd like to make it such that the order of which resources are specified is the order they're passed to the foreign proc. I think I can also make the parsing a little cleaner.

pschachte:
If we want this to be fairly general (ie, able to interface to pretty much any C function), then it's not just the order of resources that needs to be flexible.  It should also be possible to interleave the resources among the positional parameters in any order.  Also, we should probably support pass by reference for outputs.  Perhaps a Wybe in/out parameter should to translated to a single pass-by-reference C parameter.

But maybe it's better to handle the simple cases first, and extend it in later PRs as time and inspiration dictate.

jimbxb:
Interleaving can just be done in the verbose way, like:

```
def foreign c interleave(x) use !foo {
    foreign c interleave(foo, x, ?foo)
}
```

Now that I think about it, we should probably require parameters have their types specified because there's no way of inferring them :)

jimbxb:
Converting to draft while I havent made the relevant changes required.

pschachte (changes requested):
This looks great.  I just have one issue with it:  conceptually, the resources used by a proc should be thought of as a set.  In the case of foreign calls, they must be ordered, with duplication respected, as you've done.  So the WYBE.md file should mention that, in foreign definitions only, order is significant, and resource names can be repeated (and an in-out resource can be separated into an in one and an out one).  Can you please update the doc to say all that?

There's also an issue with how multiple outputs are handled.  This isn't supported by C, but we could actually handle it the way C does now, using `FlowOutByReference` arguments (ie, passing a pointer to where to write the output).  But we can leave this as future work.

One concern I do have, though, is that we continue to treat resources in non-foreign procs as if they were sets, so they should be sorted and duplicates removed, and a separated in and out resource should be coalesced into a single in-out resource.  I didn't notice that happening in the code, but I may have overlooked it.  If you haven't done that, could you please do so?

Another would-be-nice thing would be to use this functionality for all foreign functions used in the library, but again that can be a separate PR.

jimbxb:
I can do the doc updates.

On coalescing resources, I think this step might not be done yet. It should go into the normalisation phase of resources, which happens just before type checking. I'll double check and fix it up if not.

pschachte (approved):
Excellent, thanks for that!

PR #375, opened 2023-01-12
Allowed for visibility to be specified for each constructor

jimbxb:
A few other changes are also present:
* fixed visibility for some procs
* improved error messages when fields of the same name have differing visibility/type
* clean up Normalisation. It is a bit of a mess and hard to maintain


pschachte (changes requested):
Nice!  I always meant to support individual visibilities for constructors, but couldn't think of a good syntax.  This one is good.

Could you please document this in Wybe.md?

Also, could you please add a test case to ctor_visibility.wybe that has the `constructors` keyword without pub before it?  That should still be OK, but it's worth having an explicit test case.

jimbxb:
Of course... Always the WYBE.md.

pschachte (changes requested):
Also, please update WYBE.md to document the need for the `constructor` keyword in `type` declarations.

pschachte (approved):
Thank you!

PR #377, opened 2023-02-04
Incorporate parameter information into global flows

### LINKED ISSUE #376: Incorporate parameter information in global flows
Currently, the analysis of global flows for higher order calls is extremely conservative. That is, either all global's flow in and out of a higher order call or none do. This represents some issues where a global variable isn't actually modified through a higher order call.

Consider the following:

```
def {noinline} do(f:{resource}()) {
    !f
}

resource res:int = 0

?res = 1
!do {resource}{ pass }
!print res
```
This code is effectively the following in LPVM
```
def {noinline} do(f:{resource}())<univ, univ>:
    f

def anon()<{}, {}>:
    pass

foreign lpvm store(1, <<res>>)
do(anon)<univ, univ>
foreign lpvm load(<<res>>, ?res)
print(res)<{io}, {io}>
```

The `<univ, univ>` global flows are here because we don't know what global variables the resourceful HO call could touch, so we must assume all are. But, we do -- it is whatever the global flows are of the `f` parameter.

With the knowledge of the global flows of what is passed in at the call site of `do`, we should be able to remove the `load` instruction. Modifying the global flows of `do` to be `<{}, {}, {f}>` (no first order flows + flows of `f`), the call site call can be inferred to have no global flows. (This also proves that this `do` call has no effects and can be removed!)

The code would then be optimised into;
```
def {noinline} do(f:{resource}())<{}, {}, {f}>:
    f

def anon()<{}, {}, {}>:
    pass

foreign lpvm store(1, <<res>>)
print(1)<{io}, {io}, {}>
```

In essence, the global flows of a procedure would have to include the first order global flows, *and* the global flows of the arguments, if called. Special care would have to be taken when calling a non-parameter, but I believe the global flows of a local variable can be restricted to a set of parameter's global flows (or fall back to the universal set).

This analysis is only useful at the call site, not inside the body. This is still beneficial, I believe. Think at the callsite of a resourceful `map`

jimbxb:
Closes #376


jimbxb:
Oops. Gotta stop doing that

jimbxb:
My E key is almost broken. Some Es are missing from commit messages :)

src/Expansion.hs
```diff
@@ -75,22 +78,30 @@ markParamNeededness isClosure used _ param@PrimParam{primParamName=nm,
...
+    param {primParamInfo=info{
+            paramInfoUnneeded=Set.member nm ins
+                                && (not isClosure || ft /= Ordinary)}}
 markParamNeededness _ _ _ PrimParam{ primParamFlow=FlowOutByReference } =
     shouldnt "unexpected FlowOutByReference at this stage of compilation"
 markParamNeededness _ _ _ PrimParam{ primParamFlow=FlowTakeReference } =
     shouldnt "unexpected FlowTakeReference at this stage of compilation"
 
 
+-- | Update the global flows of a parameter, given the
```
pschachte, line 97: ... "given the" what?

src/Expansion.hs
```diff
@@ -75,22 +78,30 @@ markParamNeededness isClosure used _ param@PrimParam{primParamName=nm,
...
     shouldnt "unexpected FlowOutByReference at this stage of compilation"
 markParamNeededness _ _ _ PrimParam{ primParamFlow=FlowTakeReference } =
     shouldnt "unexpected FlowTakeReference at this stage of compilation"
 
 
+-- | Update the global flows of a parameter, given the 
+updateParamGlobalFlows :: Map PrimVarName GlobalFlows -> PrimParam -> PrimParam 
+updateParamGlobalFlows varFlows param@(PrimParam name _ _ _ info@(ParamInfo _ flows)) =
+    param {primParamInfo=info{
+            paramInfoGlobalFlows=flows `globalFlowsIntersection` newFlows}}
```
pschachte, line 101: Is intersection the right thing here?  A bit of explanation would be helpful.

pschachte (changes requested):

src/BodyBuilder.hs
```diff
@@ -121,20 +124,21 @@ type BodyBuilder = StateT BodyState Compiler
...
+      tmpCount       :: Int,             -- ^The next temp variable number to use
+      buildState     :: BuildState,      -- ^The fork at the bottom of this node
+      parent         :: Maybe BodyState, -- ^What comes before/above this
+      globalsLoaded  :: Map GlobalInfo PrimArg,
                                         -- ^The set of globals that we currently
                                         -- know the value of
-      reifiedConstr :: Map PrimVarName Constraint
+      varGlobalFlows :: Map PrimVarName GlobalFlows,
+      reifiedConstr  :: Map PrimVarName Constraint
                                         -- ^Constraints attached to Boolean vars
```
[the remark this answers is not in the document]
jimbxb: Will do.

src/BodyBuilder.hs
```diff
@@ -858,6 +886,43 @@ updateGlobalsLoaded prim pos = do
...
+              (\(ArgVar name ty _ _ _) ->
+                  (name, if isResourcefulHigherOrder ty
+                         then univGlobalFlows
+                         else if genericType ty
+                         then inFlows
+                         else emptyGlobalFlows)) outs
+        PrimForeign "lpvm" "load" _ [_, ArgVar name ty flow _ _]
+            | isResourcefulHigherOrder ty -> 
+          return $ Map.singleton name univGlobalFlows
+        PrimForeign{} -> do
```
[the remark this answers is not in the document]
jimbxb: That's already handled. All outputs take the union of all input's global flows. This handles `move`s and `access`es.

src/Expansion.hs
jimbxb: Good question. Will finish this off

src/Expansion.hs
jimbxb: I think intersection makes sense here. The global flows assigned initially to each term are as general as possible. We then are able to restrict the globals that *may* be interacted with to some smaller collection. 

One alternative is to just use the value from the map -- in such a case we may add to the global flows of the parameter, and then any analysis that used the old global flows may be invalidated. Same as if we used the union.

There's no way that the global flows should be added to, from what I see. We define the global flows in terms of the resources that are `use`d in the source code, and the only way a different resource could be modified is in a `use` block -- but in a `use` block, the callee never sees these effects, so we can disregard these resources. 

I used the same logic in Optimise.hs when dealing with the inference of global flows originally, using the intersection with the old and the analysed. I should probably document it, if I hadn't already.

pschachte (approved):

jimbxb:
Uh oh. I think we need to run an update-exp.

PR #381, opened 2023-04-09
Added anonymous function syntax

### LINKED ISSUE #379: Add support for anonymous functions as a syntax sugar for an anonymous proc
Currently, the anonymous procedure syntax doesn't support "functions". For example, if we wanted to write a "successor" function, we would have to do:

```
{ @ + 1 = ?@ }
```

While this isn't too long, I propose an alternative syntax for a functional style:

```
@( @ + 1 )
```

I know what you're asking: why the leading `@`? The issue I foresee without the `@` or some other symbol to delineate that this is an anonymous proc is that there's no clear point wherein you can definitely say where the function starts or ends.

Consider this entirely made up, but somewhat sensible example where we drop the `@`:

```
map(compose(@ + 1, @ * 2), foo, ?bar)
```

While reading this, it's probably clear that we are composing two functions, but I'm not sure this is super easy to parse -- why are these `@`s special. 

Perhaps you could say that the fact they are separate arguments makes it easy to distinguish, but also consider this expression:

```
(@ + 1) <| (@ * 2)  # where `<|` is a composition operator
```

To the parser/normaliser, this looks no different to the `compose` example we had before. 

I hope I've been able to make my point clear why I think it's a lot easier if we use some special syntax for functions. It's a cost of up to 3 extra characters, and would help disambiguation. 

I suppose one other cost is that we'd now have to treat `@` as a prefix operator (which is already sort of parsed like it is). Grumble.

Edit: yes, I realise I could write something like `(@ + 1) * 2`. I wanted to keep the example simple

jimbxb:
Closes #379.

Will cause problems with #377, and either of these branches will have to have `update-exp` ran before merging it. 
The changes should be compatible, though.

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -547,6 +547,16 @@ This syntax is a syntactic sugar for an anonymous procedure, that is equivalent 
 
 with the output `@` with an appropriate numbering. 
 
+For example, the following pairs are equivalent:
+
+```
+@( @ + 1 )
+{ ?@2 = @1 + 1 }
+
+@( @2 ^ @1 )
+{ @2 ^ @1 = ?@3 }
+```
+
 
 ##  <a name="operator-syntax"></a>Operator syntax
 
--- a/WYBE.md
+++ b/WYBE.md
@@ -531,6 +531,22 @@ f(2, ?j) # j = 3; i = 1
 f(3, ?j) # j = 4; i = 2
 ```
 
+### Anonymous functions
+
+Anonymous functions are also supported with the following syntax:
+
+> `@(` *exp* `)`
+
+Where *exp* is an expression that may contain a number of `@` holes as
+per the parameter syntax of anonymous procedures. These parameters must all
+have an input mode, and the expression may only contain input-mode variables.
+
+This syntax is a syntactic sugar for an anonymous procedure, that is equivalent to the following:
+
+> `{ foreign llvm move(` *exp*`, ?@) }`
+
+with the output `@` with an appropriate numbering. 
+
 
 ##  <a name="operator-syntax"></a>Operator syntax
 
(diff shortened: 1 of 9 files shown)


test-cases/final-dump/higher_order_anon.wybe
```diff
@@ -0,0 +1,13 @@
+?incr = { 1 + @ = ?@ }
+?incr = @(1 + @)
+
+?const = { foreign llvm move(@2, ?@3) }
+?const = @(@2)
```
pschachte, line 5: Shouldn't that be `@(@1)`?  I'd expect `const` to take two arguments and return the first.

WYBE.md
```diff
@@ -531,6 +531,22 @@ f(2, ?j) # j = 3; i = 1
...
+Where *exp* is an expression that may contain a number of `@` holes as
+per the parameter syntax of anonymous procedures. These parameters must all
+have an input mode, and the expression may only contain input-mode variables.
+
+This syntax is a syntactic sugar for an anonymous procedure, that is equivalent to the following:
+
+> `{ foreign llvm move(` *exp*`, ?@) }`
+
+with the output `@` with an appropriate numbering. 
+
```
pschachte, line 549: An example anonymous expression with its equivalent anonymous proc would be helpful here.

pschachte (approved):
You can merge this when you're ready.

test-cases/final-dump/higher_order_anon.wybe
jimbxb: Yep. Youre right. I think that might be a limitation of this syntax, actually, as I cant think of an expression that "ignores" is the second argument...

jimbxb: I just changed the example to be "ignore" (whatever `flip const` is)

PR #382, opened 2023-04-30
Add a simple interface for using randomness

jimbxb:


pschachte (approved):
This looks great, James, but could you add a test case for it before merging it?

jimbxb:
I am not sure I can add a test case for this. Randomness would probably not be good for a deterministic testing framework. 

I suppose I can make it really dumb :)

pschachte (changes requested):
Ug.  Now that I see that test case, I think it's problematic to add a proc called "generate" to the standard library.  First, that's a name programmers are likely to want to use themselves, but more importantly, it's not at all indicative of what it does.  Could you rename it to `random`, and rename `seed` to `set_random_seed`?

PR #391, opened 2023-06-02
Add JSON parser example

jimbxb:
Ideally I'd use a resource to handle the parse string, but I found a few compile errors along the way regarding uninitialised temporary variables. I didn't get a change to debug where these were, though, as they didn't have any location attached to them.

I believe one benefit for using a resource is that, for a test proc, if the proc fails there is already a mechanism to rollback the initial state for the resource. This would clean up the `char` and `digit` procs. Of course, these procs should work without the hack I built in, but this is a known issue which is to be documented on GitHub.

Overall, though, I think the use of higher order in both parsing list-like structures and in printing is a treat to work with!


samples/json.wybe
```diff
@@ -0,0 +1,251 @@
...
+        char(c, !s)
+    }
+    space(!s)
+} 
+
+def space(!s:string) {
+    if { { char(' ', !s) | char('\n', !s) | char('\r', !s) | char('\t', !s) } :: space(!s) }
+}
+
+def {test} json(!s:string, ?j:_) {
```
pschachte, line 34: I suggest being consistent about which argument, first or last, is the string being parsed.  Here it's the first, for `char` it's the last, and in some places it's in the middle.

samples/json.wybe
```diff
@@ -0,0 +1,251 @@
...
+def {test} digits(!s:string, ?len:float, !n:float) {
+    digit(!s, ?d)
+    ?n = n * 10.0 + d
+    digits(!s, ?len, !n)
+    !len += 1.0
+  | ?len = 1.0
+}
+
+def {test} digit(!s:string, ?d:float) {
+    [?c | ?rest] = s
```
pschachte, line 162: How about writing another mode for `char` (where the second argument is output) that does exactly this?  That would make the code simpler and let you avoid a leaky abstraction.

pschachte (changes requested):
This is really lovely!  Very readable code, and a nice sales pitch for Wybe.  I've made a few style nitpicks here, and also, a bit of documentation would be good for this to be a Wybe code sample.

I agree about using a resource.  It'd make a nice example use of resources, too.

jimbxb:
Thanks for the feedback though! I've been working on a version with resources that has some different names and some little tweaks to make it ergonomic, but I should add a commit with these changes before tackling the resource version.

Unfortunately, the resource version has uncovered a bad bug that I introduced with HO resources, specifically in the code generation of tests in unbranching and resource transformation. The bug is that, in a test proc, where i save the resources to be reset upon the call failing, I fail to add the tmp variables to the vardicts of nested conditionals -- I do add them to the outer most condional that's generated though. This then results in some tmp variables being "uninitialised" when it comes time to unbranch because they're not known in the vardicts.

The fix should be to just add them, but I think the Haskell code becomes very clunky then, and I'm sure I could do better. The "better" though, is probably a rework of unbranching as a whole, calculating the vardicts when unbranching. I think this is a better design as, if we ever add in more code that generates more tmp variables, then we don't have to worry about updating the vardicts after mode checking. I realise this was done in mode checking because it's convenient, but I think that in unbranching is more appropriate as that's where the data is used.

samples/json.wybe
```diff
@@ -0,0 +1,251 @@
+pub constructors jobject(list(pair)) | jstring(string) | jbool(bool) | jnumber(float) | jlist(list(_)) | jnull
+    
+pub def {test} (a:_ = b:_) { fail }
```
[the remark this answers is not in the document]
jimbxb: This is to fix an annoying bug -- there's no `=` for generic types, so the auto generated `=` here fails in type checking

samples/json.wybe
jimbxb: I was trying to follow the order of <inputs> <inouts> <outputs>. No more, now that resources are in place!

samples/json.wybe
jimbxb: Event better, I can use an implied mode of this new `char` for the existing one!

jimbxb:
I finally got back into things...

The parser now "works", but only if I turn off multi-specz. WIth multi-specz on I get a segfault!

test-cases/complex/exp/testcase_multi_specz-drone.exp
```diff
@@ -99,9 +99,8 @@ module top-level code > public {semipure} (0 calls)
 ()<{<<command_line.arguments>>, <<command_line.command>>}; {<<command_line.arguments>>, <<command_line.command>>}; {}>:
   AliasPairs: []
   InterestingCallProperties: []
-  MultiSpeczDepInfo: [(0,(wybe.array.[|]<0>,fromList [NonAliasedParamCond 2 []]))]
     foreign lpvm load(<<command_line.arguments>>:wybe.array(wybe.c_string), ?%arguments##0:wybe.array(wybe.c_string))
-    wybe.array.[|]<0>[785a827a1b](?command##1:wybe.c_string, ?arguments##2:wybe.array(wybe.c_string), ~arguments##0:wybe.array(wybe.c_string), ?tmp#8##0:wybe.bool) #0 @command_line:nn:nn
+    wybe.array.[|]<0>(?command##1:wybe.c_string, ?arguments##2:wybe.array(wybe.c_string), ~arguments##0:wybe.array(wybe.c_string), ?tmp#8##0:wybe.bool) #0 @command_line:nn:nn
```
jimbxb, line 103: This is an unfortunate side-effect, but we shouldnt just be mutating a global variable like this! :(

jimbxb:
Next step, Wybe quines!

test-cases/complex/exp/testcase_multi_specz-drone.exp
jimbxb: I suppose we can add a NonAliasedParamCond for globals. That way, if we know the global _isnt_ aliased, a load of the global wouldn't load an aliased value, and we can optimise accordingly. 

This is tricky, as if we ever store into that same global, the alias information changes to whatever the stored value had. 

Recognising that a global is unaliased feels tricky too, but I assume you can just feed in the aliasing information of whatever you store into the global.

Food for thought, but probably not something I'm game to implement.

samples/json.wybe
```diff
@@ -0,0 +1,277 @@
...
+  | !list(?j)
+  | !null & ?j = jnull
+}
+
+## sequence/5
+# Parse a comma-separated sequence of things, with a preceding and 
+# following character, followed by whitespace.
+def {test} sequence(before:char, parse_item:{resource,test}(?X), ?xs:list(X), after:char) use !tokens {
+    !literal(before)
+    if { parse_item(?x) ::
```
jimbxb, line 78: This should have a !. Not sure why there's no error.

samples/json.wybe
```diff
@@ -0,0 +1,277 @@
...
+    for ?c in str {
+        !char(c)
+    }
+    !space
+}
+
+## space/1
+# Parse zero or more whitespace characters from the input.
+def space() use !tokens {
+    !char(?c) & (c = ' ' | c = '\n' | c = '\r' | c = '\t')
```
jimbxb, line 58: Oops. This only handles one whitespace. Need to recurse

samples/json.wybe
```diff
@@ -0,0 +1,277 @@
...
+           | c = 'b' :: ?c = '\b'
+           | c = 'f' :: ?c = '\f'
+           | c = 'n' :: ?c = '\n'
+           | c = 'r' :: ?c = '\r'
+           | c = 't' :: ?c = '\t'
+           | else :: fail
+        }
+    }
+    !string_tail(?str:string)
+    # this should work...
```
jimbxb, line 139: I do want to fix this. Cons for strings was conflicting with lists

samples/json.wybe
```diff
@@ -0,0 +1,277 @@
...
+}
+
+## digits/3
+# Parse a sequence of digits, outputting the number of digits parsed
+def {test} digits(?len:float, !n:float) use !tokens {
+    !digit(?d)
+    ?n = n * 10.0 + d
+    !digits(?len, !n)
+    !len += 1.0
+  | ?len = 1.0
```
jimbxb, line 179: This should be 0.0.

jimbxb: I just need to +1.0 elsewhere

test-cases/complex/exp/testcase_multi_specz-drone.exp
pschachte: Yes, very tricky.  We need a global analysis to work out that we never either store a potentially aliased value into, and we never alias a value we load from, this global.  Alternatively (or additionally), we need a way to declare that a resource should never be aliased, and always generate code that duplicates the value after loading if it can't be shown that the stored value will not be used again, and before storing if it can't be shown that the stored value is unaliased.  Something like that.  This is a whole other can of worms, certainly not related to a json parser.

pschachte (changes requested):
Awesome!  This looks really nice, just the way I hoped a Wybe parser could look.
And thanks for fixing many issues in the compiler.

Just one request:  could you add a header comment to the top of `json.wybe` saying that it's an example of how one might write a recursive-descent parser for JSON documents in Wybe?

pschachte (approved):
Great; thanks for fixing the Brainfuck doc, too!

PR #396, opened 2023-08-08
Fix unbranching issue with in-out params that are unused

jimbxb:
@wnats


wnats:
@pschachte This is a fix to the "Schrödinger" problem I mentioned in one of the recent meetings, where the length of an array seems to be in a "superposition".

pschachte (approved):
Thanks, for that, James!

PR #401, opened 2023-10-02
Fixed executable module importing mod dependencies not in topological order

wnats:
`a.wybe`:
```
pub resource a_res:int = 2:int
```
`b.wybe`:
```
use a  # doesn't matter if it's pub
?b = 0  # anything as long as it's a top-level stmt
```
`c.wybe`:
```
use a,b
?c = 0
```

Running `wybemk c` with the above files would yield the following error message:
```
Error detected during type checking of module(s)
Input resource a_res not available at call to proc
```

The logs reveal that the main executable module imports the module dependencies in a non-topological order (`[b, a, c]` instead of `[a, b, c]`) (credit to @jimbxb):
```
 Module 
  representation  : (not a type)
  public submods  : 
  public resources: 
  public procs    : 
  imports         : use a
                    use b
                    use c
                    use command_line
                    use wybe
                    use wybe.array
                    use wybe.bool
                    use wybe.c_string
                    use wybe.char
                    use wybe.comparison
                    use wybe.control
                    use wybe.count
                    use wybe.float
                    use wybe.int
                    use wybe.io
                    use wybe.list
                    use wybe.machine_word
                    use wybe.memory_management
                    use wybe.phantom
                    use wybe.predicate
                    use wybe.range
                    use wybe.string
  resources       : 
  procs           : 

module top-level code > {terminal,impure} (0 calls)
0: () use command_line.argc, command_line.argv, ?command_line.exit_code:
    use a.a_res, command_line.arguments, command_line.command, wybe.io.io in
        !wybe.memory_management.
        !wybe.io.
        !command_line.
        !b.
        !a.
        !c.
        foreign c {semipure,terminal} exit(exit_code:wybe.int)
    }

LLVM code       : None
```

This is caused by `orderedDependencies` in `Builder.hs` collecting the imported modules in a purely DFS fashion, which is not necessarily topological.
https://github.com/pschachte/wybe/blob/54f51c55ad9ba5fd583ffebd5d0aa0e12539b197/src/Builder.hs#L1160-L1197

This PR fixes this problem by stably sorting some of the collected modules on their (topological) order in `orderedSCCs`.

### DIFF
--- a/src/Builder.hs
+++ b/src/Builder.hs
@@ -233,7 +233,7 @@ import           Transform                 (transformProc,
 import           Types                     (typeCheckModSCC,
                                             validateModExportTypes)
 import           Unbranch                  (unbranchProc)
-import           Util                      (sccElts, useLocalCacheFileIfPossible)
+import           Util                      (sccElts, useLocalCacheFileIfPossible, (&&&))
 import           Snippets
 import           Text.Parsec.Error
 import           BinaryFactory
@@ -935,6 +935,10 @@ isStdLib :: ModSpec -> Bool
 isStdLib []    = False
 isStdLib (m:_) = m == "wybe"
 
+-- | Filter for avoiding the command line module, which is currently hardcoded
+--   into the executable module
+isCmdLine :: ModSpec -> Bool
+isCmdLine = (==["command_line"])
 
 -- |A Processor processes the specified module one iteration in a
 --  context of mutual dependency among the list of modules.  It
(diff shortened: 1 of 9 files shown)


wnats:
Sorry the fix is missing something minor, will reopen soon

src/Builder.hs
```diff
@@ -935,6 +935,10 @@ isStdLib :: ModSpec -> Bool
 isStdLib []    = False
 isStdLib (m:_) = m == "wybe"
 
+-- | Filter for avoiding the command line module, which is currently hardcoded
+--   into the executable module
+isCmdLine :: ModSpec -> Bool
+isCmdLine = (==["command_line"])
```
jimbxb, line 941: Can we factor this into a constant? It's used in a few places now :)

jimbxb (changes requested):

src/Builder.hs
wnats: Would `Snippets.hs` be a good place?

jimbxb: Possibly. I forget where these constants are meant to go :(

src/Builder.hs
```diff
@@ -1029,7 +1029,9 @@ buildExecutable orderedSCCs targetMod target = do
     loadModuleIfNeeded False ["command_line"] possDirs
     let privateImport = importSpec Nothing Private
     addImport ["command_line"] privateImport `inModule` targetMod
```
jimbxb, line 1031: Can we use the constant you just added here :) please check for other references to `"command_line"` too if you can

jimbxb (changes requested):

jimbxb (approved):
Great

pschachte (changes requested):
This looks great, but can you please also add your test case to the test suite, probably as a final dump test, so we can catch any regression on this bug fix?

Thanks!

wnats:
@pschachte I'm adding the tests as execution tests since final dump doesn't build an executable module, which this bug is concerned about. I added a distinct print statement on each file to show the order every module's main proc is called.

wnats:
Actually, instead of having three files, I've just realised I could just have two submodules in the driver file. Will do so before I reopen the PR.

pschachte (approved):
Great, thanks!

Issue #125, opened 2021-03-29
Develop a relational type system


pschachte:
The Wybe type system is functional, in that constructors are invertible injective functions.  This has the limitation of only being able to construct directed acyclic graphs.  It also does not support a concept of object identity.

A relational type system would be more like a relational database, in that objects could be related to one another, and it would be possible to navigate between them bidirectionally.  Also, such a system would support one-one, one-many, and many-many relations.  This would allow entity types (ie, identities) to be declared, specifying their attributes, and would allow relations among entities to be declared, explicitly specifying whether they are one-one, one-many, or many-many.  Other multiplicities may be useful, such as one-to-one-or-more.

Entities would live in an explicit store (which would probably treated as a resource).  Entities would be created by specifying all their attributes (which can be any Wybe value) and all their mandatory related entities.  Attributes could then be updated, and relationships could be changed, producing a new store.

The implementation should be maximally efficient.  Entities should be represented as data structures holding the entity's attributes and references to related entities.  Hashed indices should be transparently generated and maintained for efficiency. As long as the store is unique (linearly threaded through the code), updates to attributes and relations should be implemented through destructive update.  In cases where the store is not unique, implementation becomes more difficult.

One benefit of this project is that it would make it possible to translate imperative algorithms and data structures based on destructive update to Wybe, and produce similar efficiency.  It would also allow a more natural translation from a data model (eg, a UML class diagram) to Wybe code.  For example, a system with subjects and students could allow a student to enrol or unenrol in a subject, and make it simple to find all the students in a given subject and all the subjects a given student is enrolled in.

In addition to relational algebra, it's worth looking at [Fact-Based Modelling](http://factbasedmodeling.org/) for inspiration.

PR #404, opened 2023-10-04
Delayed flattening step from preliminary processing to final normalisation

wnats:
Previously, the flattening (of proc bodies) step occurred as the modules were loaded top-down. However, with future enhancements, there may be statements that require some knowledge of modules that have yet to be collected.

For example, flattening an entity lookup statement (#125) would require knowing which attributes are indexed for optimisation purposes, which requires the entity module to be loaded.

A potential alternative would be to delay flattening these statements, but @pschachte suggested just moving the flattening step to the final normalisation stage.

I have also updated one final dump test, which catches an error triggered during flattening. Since the flattening step has moved, the error message differs slightly, specifically at which stage the error occurred.


src/Flatten.hs
```diff
@@ -57,25 +57,29 @@ import Control.Monad.Trans (lift,liftIO)
...
-    (stmts',tmpCtr) <- flattenBody stmts (inParams `Set.union` resources)
-                       (modifierDetism mods)
-    return (ProcDecl vis mods proto stmts' pos,tmpCtr)
-flattenProcDecl _ =
-    shouldnt "flattening a non-proc item"
+        ProcDefSrc body = procImpln pd
+        detism = procDetism pd
+
+    logMsg Flatten
+        $ "** Flattening " ++ "def " ++ show proto
```
jimbxb, line 75: Does this need to show the mods too, as before?

Could also manually append the two constant strings :)

src/Flatten.hs
```diff
@@ -57,25 +57,29 @@ import Control.Monad.Trans (lift,liftIO)
...
+            ++ " {" ++ showBody 4 body ++ "}"
+    mapM_ (placedApply $ flip explicitTypeSpecificationWarning . paramType) params
```
jimbxb, line 77: Does this need to be `flip`ped? Can we flip the definition instead if it's not used in the normal order?

src/Flatten.hs
wnats: Good point, I have just pushed to reflect on this.

src/Flatten.hs
wnats: Yeah, I didn't really like how it looked tbh. It's used in other parts of the code without a `flip` or an infix notation, unfortunately. :\

jimbxb: Oh well!

jimbxb (approved):

pschachte (changes requested):
Looks good, aside from one request.

src/Builder.hs
```diff
@@ -794,6 +795,7 @@ compileModSCC mspecs = do
     -- FLATTENING
     fixpointProcessSCC handleModImports mspecs
     completeNormalisation mspecs
+    mapM_ (transformModuleProcs flattenProcBody) mspecs
```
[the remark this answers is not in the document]
wnats: `completeNormalisation` is in `Normalise.hs`, while `transformModuleProcs` is in `Builder.hs`, so if I put the former I would have to import `transformModuleProcs` from `Normalise.hs`, which I'm unsure about as `Builder.hs` has already imported from `Normalise.hs`.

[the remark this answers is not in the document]
pschachte: Yeah, circular imports are a PITA in Haskell.

You could move `transformModuleProcs` into AST.  That's imported by both Builder and Normalise, and I don't see anything in the definition of `transformModuleProcs` that's defined in Builder, so that should fix it.  You'd just have to remove the two logBuilder lines from the definition, which is fine because reenterModule and reexitModule already log what they're doing.

pschachte (approved):
Thank you!

PR #419, opened 2023-11-09
Revert on failure

### LINKED ISSUE #411: After a failing test, variables are not backtracked to their values before the test
Here's an example:
```
def {test,noinline} nope { fail }

def yup use !io {
    ?x = 42
    if { ?x = 24 ; nope :: !println("nope succeeds: x = $x")
       | else :: !println("nope fails; x = $x")
       }
    !println("after conditional; x = $x")
}

!yup
```
This prints
```
nope fails; x = 24
after conditional; x = 24
```
when it should print
```
nope fails; x = 42
after conditional; x = 42
```

### LINKED ISSUE #412: Foreign calls with Boolean outputs can't be de-reified into tests
For example
```
pub def {test} lt(x:int, y:int) { foreign llvm icmp_slt(x,y) }
```
gets this error message
```
Foreign call 'icmp_slt' with arity 2; should be 3
```

pschachte:
Fix #411 
Fix #412 
Much code refactoring, especially in Types and Unbrancher modules, mainly to move information passed as arguments into state monads.


foo.wybe
```diff
@@ -0,0 +1,37 @@
+def {test,noinline} nope { fail }
```
jimbxb, line 1: We probably want to remove these :)

jimbxb (changes requested):

PR #423, opened 2023-11-20
Added Benchmarking Library

wnats:
Added a library for benchmarking purposes, as requested by @pschachte.

This is similar to #304, but with @pschachte's suggestion to store the clock counter as a global variable in `cbits.c`.

This library introduces two new `semipure` procedures, `!benchmark.start` and `!benchmark.end(?time:float)`, serving as a pair of probes to measure the _CPU clock time_ elapsed between two points of a program. A utility procedure `!benchmark.time_execution(proc:{semipure}(), ?time:float)` is also introduced which times `proc`. The library will throw an error if:
- `!benchmark.start` is called twice without a `!benchmark.end(?time:float)` in between
- `!benchmark.end(?time:float)` is called when an un-ended `!benchmark.start` has not been called
- `!benchmark.time_execution(...)` is called when there is a preceding `!benchmark.start` which has not been ended


jimbxb:
This is cool. I like that we can guarantee that the timer is only started and ended once in succession.

One request: can we add a utility of some kind like this?

```
def {semipure} time(proc:{semipure}(), ?time:float) {
    !start
    !proc
    !end(?time)
}
```

wybelibs/benchmark.wybe
```diff
@@ -0,0 +1,27 @@
...
+# End the benchmark clock, yielding the elapsed time in seconds
+# Can only be called after a start() has been called
+# which has not been ended by an end()
+pub def {semipure} end(?time:float) use call_source_location {
+    foreign c {impure} benchmark_end(call_source_location, ?time)
+}
+
+# Measure the running time of a semipure niladic procedure
+# Assumes any start() call has been ended
+pub def {semipure} time(proc:{semipure}(), ?time:float) {
```
jimbxb, line 23: Now that I think about it, we probably want the `call_source_location` from wherever we call `time`. I think that means we have to manually call the `foreign` procs in here too.

wybelibs/benchmark.wybe
```diff
@@ -0,0 +1,27 @@
...
+use wybe.c_string, wybe.float
+
+# Start the benchmark clock
+# Can only be called once before being ended by end()
+pub def {semipure} start() use call_source_location {
+    foreign c {impure} benchmark_start(call_source_location)
+}
+
+# End the benchmark clock, yielding the elapsed time in seconds
+# Can only be called after a start() has been called
```
jimbxb, line 15: Nit: can we remove the prepositions on `start` and `end`. `start` doesn't use the preposition :)

jimbxb (changes requested):

wybelibs/benchmark.wybe
wnats: Good point!

wybelibs/benchmark.wybe
wnats: No worries, didn't realise that, thank for pointing it out.

jimbxb (approved):
LGTM!

wybelibs/benchmark.wybe
```diff
@@ -0,0 +1,27 @@
...
+# End the benchmark clock, yielding the elapsed time in seconds
+# Can only be called after start() has been called
+# which has not been ended by end()
+pub def {semipure} end(?time:float) use call_source_location {
+    foreign c {impure} benchmark_end(call_source_location, ?time)
+}
+
+# Measure the running time of a semipure niladic procedure
+# Assumes the preceding start() (if any) has been ended
+pub def {semipure} time(proc:{semipure}(), ?time:float) use call_source_location {
```
pschachte, line 23: Could you rename this proc to `time_execution`?  I think this name is pretty ambiguous.

pschachte (changes requested):
All good except one request.

pschachte (approved):

pschachte:
Thank you!

PR #427, opened 2023-12-08
Fixed string concat to cstring conversion

wnats:
```
?str = ("0" ,, "1") ,, "2"

!println(c_string(str))
```
would incorrectly output `"01"`, at least in my local machine which at that time happened to allocate a zeroed out memory chunk on an `lpvm alloc`. This problem is more noticeable when calling `!logmsg(<log_str>)`, as `<log_str>` is first converted into a cstring.

The culprit seems to be the `incr(!offset)` line. The line is unnecessary as `offset` has been incremented accordingly in each recursive `pack` call.
https://github.com/pschachte/wybe/blob/35b5c146366c6a4693ce4c81a3a821933750fe7c/wybelibs/wybe/string.wybe#L180-L183

Perhaps unrelated to the main problem, but I have also changed the third argument of `mutate` to `len - 1`, which I believe is more accurate, although I have yet to see a corner case for this.
https://github.com/pschachte/wybe/blob/35b5c146366c6a4693ce4c81a3a821933750fe7c/wybelibs/wybe/string.wybe#L40-L49


jimbxb:
LGTM. This is a good catch!

jimbxb (approved):

pschachte (approved):
Thanks for fixing this.

PR #436, opened 2024-01-06
Add error function

### LINKED ISSUE #434: Add an `error` function returning `any` type
This is would resemble Haskell's `error` which matches with any type. One motivation would be in conditional/case _expressions_ where we have exhaustively list all possible outcomes but the compiler still requires an `else` branch. For instance, the following example in `WYBE.md` would work:
https://github.com/pschachte/wybe/blob/29ae3cbf675fde812066477da17ca45c3a277ad6/WYBE.md?plain=1#L1169-L1176

One primary challenge would be handling the special resource `call_source_location` which the current `error` _statement_ uses.

Although of course, a better long-term solution would be #19.

pschachte:
Closes #434

### DIFF
--- a/wybelibs/wybe/control.wybe
+++ b/wybelibs/wybe/control.wybe
@@ -18,6 +18,10 @@ pub def {terminal,semipure} error(message:string) use call_source_location {
     foreign c {terminal,semipure} error_exit(call_source_location, c_string(message))
 }
 
+pub def {terminal,semipure} error(message:string, result:T) use call_source_location {
+    !error(message)
+}
+
 pub def {terminal,semipure} error(message:c_string) use call_source_location {
     foreign c {terminal,semipure} error_exit(call_source_location, message)
 }
--- a/wybelibs/wybe/control.wybe
+++ b/wybelibs/wybe/control.wybe
@@ -18,7 +18,7 @@ pub def {terminal,semipure} error(message:string) use call_source_location {
     foreign c {terminal,semipure} error_exit(call_source_location, c_string(message))
 }
 
-pub def {terminal,semipure} error(message:string, result:T) use call_source_location {
+pub def {terminal,pure} error(message:string, ?result:T) use call_source_location {
     !error(message)
 }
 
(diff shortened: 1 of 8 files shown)


jimbxb:
Just want to check: the `call_source_location` isn't passed into `error/1` is it? Meaning the resource usage here is redundant.

wybelibs/wybe/control.wybe
```diff
@@ -18,6 +18,10 @@ pub def {terminal,semipure} error(message:string) use call_source_location {
     foreign c {terminal,semipure} error_exit(call_source_location, c_string(message))
 }
 
+pub def {terminal,semipure} error(message:string, result:T) use call_source_location {
```
jimbxb, line 21: `result` should have `?` mode

jimbxb (changes requested):

pschachte:
Yes, the source position is passed from the error function to the error proc, so that the source position is the position of the call to the error function.  So the `use` is needed.  But you're quite right that `result` needs a `?` flow.  I've added a test case to verify, and I had to allow calls to procs with only implicit resources to omit the `!` to allow function calls to have implicit resources.

jimbxb:
Oh, I wasn't aware of that functionality.

When the special resources are threaded though the program, it was my understanding that something like

```
def foo use call_source_location {
    !bar
}

def bar use call_source_location {
    pass
}
```
would become

```
def foo(call_source_location: string) {
    foreign lock move("location of bar call in foo", ?tmp)
    !bar(tmp)
}

def bar(call_source_location: string) {
    pass
}
```
with the resource not being passed from the `foo` caller into `bar`, but instead the location of the call in `foo` is passed through to `bar`. 

I think the expected behaviour is as described here too, as the implicit resources contain information about the call's source location, not the parent's source location.

pschachte:
In this example, `foo` would pass its parameter to `bar`.  The call to `foo` pass the source location of the call as an argument in the call.  So the actual source location is the location of the call to the proc that `use`s an implicit resource from a proc that doesn't `use` that resource.  Which is what you want so the error function can call the error proc, and have the location of the call to the error function printed out.

PR #439, opened 2024-01-15
Only save/restore variables bound before a cond/disj

wnats:
Compiling `a.wybe`
```
(
    ?a = 0
    !a += 1
|
    pass
)
```
would yield
```
Error detected during generating low level code in a
Uninitialised variable 'a'
```
This is caused by `saveRestore` in `Types.hs` simply saving/restoring variables that are reassigned in a disjunction branch, including those that are unbound before the disjunction.
https://github.com/pschachte/wybe/blob/29ae3cbf675fde812066477da17ca45c3a277ad6/src/Types.hs#L2495-L2498

Same issue for a conditional e.g.
```
def {test, noinline} suc { pass }
# The suc is added to make the cond tests not trivially deterministic
if {
    ?a = 0; suc; !a += 1 ::
        pass
|   else ::
        pass
}
```


pschachte (approved):
Looks good, thanks for that!

jimbxb (approved):

PR #448, opened 2024-02-17
Fixed #447

### LINKED ISSUE #447: Current module alias in resource initialisation is not yet resolved when building the main executable
Compiling
```
constructor ctor

resource res:_ = foreign lpvm cast(0):_
pass
```
would yield
```
Prelude.init: empty list
```
This is caused by `init currMod` when looking up the type of `foreign lpvm cast(0)`:
https://github.com/pschachte/wybe/blob/12deaa6bbf124fcbe7e02c21eebb80cecbddca66/src/AST.hs#L952-L955

This `lookupType'` call happened during the building of the main executable, since #426 moved the resource initialisation process there. The currently loaded module at that time hence was the main executable module.

wnats:
See #447


pschachte (changes requested):
This looks great; thanks for fixing it.  But can you please add an execution test instead of a final_dump test, so we know if it builds a correct main?  You should probably add a bit of code that actually uses the resource.

pschachte (approved):
Excellent, thanks.

jimbxb (approved):

Issue #452, opened 2024-04-04
Floating point formatting error


pschachte:
!println "the answer is $(66.6/1.8)"
prints:
`the answer is 36.:`
when it should print:
`the answer is 37.0`
(note that : is the next character after 9).
This is a bug in the definition of float:fmt.

pschachte:
Fixed by  #453

PR #453, opened 2024-04-04
 Fix bug #452

pschachte:


jimbxb:
It would be nice to have an overloaded version of fmt that you can pass the precision into (probably an int, for number of decimal places)

Issue #454, opened 2024-04-07
Add overloaded versions of fmt for the float type


pschachte:
There should be versions of float.fmt with extra arguments to specify precision and whether to use exponential notation.

Issue #455, opened 2024-04-07
Generalise string interpolation


pschachte:
There should be a string interpolation syntax for specifying minimum field size and left/right/centre justification.  This should be independent of the type of item being formatted, so it should not require any change to the fmt functions.

pschachte:
Closed by #458

PR #458, opened 2024-04-16
Generalise fmt function to support field with and fill char

pschachte:


pschachte:
Closes #454 #455
This doesn't actually allow justification to be specified, but does allow fill character to be specified.  But specifying justification for numeric fields is probably not really desirable, so I'll close that request, too.

PR #471, opened 2024-10-08
Update compiler to use latest LLVM version

pschachte:
Unfortunately, it does this by generating a separate .ll file and explicitly compiling it with llc.


src/AST.hs
```diff
@@ -1592,15 +1599,15 @@ data ModuleImplementation = ModuleImplementation {
                                               -- ^Resources visible to this mod
     modKnownProcs:: Map Ident (Set ProcSpec), -- ^Procs visible to this module
     modForeignObjects:: Set FilePath,         -- ^Foreign object files used
-    modForeignLibs:: Set String,              -- ^Foreign libraries used
-    modLLVM :: Maybe LLVMAST.Module           -- ^Module's LLVM representation
+    modForeignLibs:: Set String               -- ^Foreign libraries used
+    -- modLLVM :: Maybe LLVMAST.Module           -- ^Module's LLVM representation
```
jimbxb, line 1603: Can we remove these that have been commented out?

src/AST.hs
```diff
@@ -3363,11 +3440,28 @@ argIsConst ArgFloat{}          = True
 argIsConst ArgString{}         = True
 argIsConst ArgChar{}           = True
 argIsConst (ArgClosure _ as _) = all argIsConst as
+-- XXX Should we consider globals to be constants, since they are compile-time
```
jimbxb, line 3443: This is interesting. They should only ever occur in a `load` or `store` instruction, and they only ever load/store the global, which is a constant reference.

If we were to ever generate code for a "global" argument, would we dereference the global or pass a reference to the global (the address) -- this would determine if it's a constant or not.

test-cases/execution/type_generics.exp
```diff
@@ -4,7 +4,5 @@
 1
 1
 1
-1.000000
```
jimbxb, line 7: Should these be removed?

pschachte: Certainly not!  Thanks for catching this.

src/AST.hs
pschachte: Done.

src/AST.hs
pschachte: Looking at places where this function is used, the argument only needs to be fixed value at compile time.  Updated doc to clarify this.

logging.ll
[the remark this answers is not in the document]
pschachte: Yep.  It was useful keeping them around while I was developing, and then I forgot to cleanup when I was finished.  Fixed now.

src/AliasAnalysis.hs
```diff
@@ -342,8 +342,8 @@ _maybeAliasPrimArgs args = do
             -> do
                 isPhantom <- argIsPhantom arg
                 rep <- lookupTypeRepresentation ty
-                -- Only Address type will create alias
-                if not isPhantom && rep == Just Address
+                -- Only Pointer type will create alias
+                if not isPhantom && rep == Just Pointer
```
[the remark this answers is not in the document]
pschachte: Actually, I think it can, kind of.  If someone tries to make an copy of the io state, that would be bad, and we do track that.  But in that case, maybe we want to do this test regardless of whether it's a phantom.  I'll try taking this out and see what happens.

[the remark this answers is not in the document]
jimbxb: But io isn't a pointer, and I hope it never could be. Phantoms must be zero-width, and pointers (I'd hope!) should be word-sized.

src/AST.hs
pschachte: Done

src/AliasAnalysis.hs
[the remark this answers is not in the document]
pschachte: Good point.
I've removed that test, and the test that io can't be copied still passes.

jimbxb:
Oops. I didn't mean to change that.

jimbxb:
I am running an Ubuntu install right now and thought I would give this branch a spin because I am keen to see its completion

So I did some tinkering... I found that basically all of the tests fail, just like with the GH worker.

I have found that swapping out the `__LPVM.__lpvm` with the MacOS `__LLPM,__lpvm` section name seem to get the final dumps working!

Still, all of the execution tests fail. This is why...

```
❯ ./wybemk -Lwybelibs test-cases/execution/string-head           
cc: error: : No such file or directory

cc: error: : No such file or directory
```

Taking the dump of what emit logs I find this at the end:

```
Emit: Generating final executable with command line: cc -no-pie -Wl,--gc-sections /tmp/wybe-ed4d53a612eeda26/tmpMain.o  /home/james/docs/wybe/test-cases/execution/string-head.o /home/james/docs/wybe/wybelibs/command_line.o /home/james/docs/wybe/wybelibs/wybe/_.o /home/james/docs/wybe/wybelibs/wybe/array.o /home/james/docs/wybe/wybelibs/wybe/bool.o /home/james/docs/wybe/wybelibs/wybe/c_string.o /home/james/docs/wybe/wybelibs/wybe/char.o /home/james/docs/wybe/wybelibs/wybe/comparison.o /home/james/docs/wybe/wybelibs/wybe/control.o /home/james/docs/wybe/wybelibs/wybe/count.o /home/james/docs/wybe/wybelibs/wybe/float.o /home/james/docs/wybe/wybelibs/wybe/int.o /home/james/docs/wybe/wybelibs/wybe/io.o /home/james/docs/wybe/wybelibs/wybe/list.o /home/james/docs/wybe/wybelibs/wybe/machine_word.o /home/james/docs/wybe/wybelibs/wybe/memory_management.o /home/james/docs/wybe/wybelibs/wybe/opaque_pointer.o /home/james/docs/wybe/wybelibs/wybe/phantom.o /home/james/docs/wybe/wybelibs/wybe/predicate.o /home/james/docs/wybe/wybelibs/wybe/range.o /home/james/docs/wybe/wybelibs/wybe/string.o /home/james/docs/wybe/wybelibs/wybe/cbits.o -lgc -lm -o /home/james/docs/wybe/test-cases/execution/string-head
```

I cant run that command in my terminal because the tmp files get removed, unforunately. But its strange. I will continue to see if I can find out whats wrong later this weeken

jimbxb:
Turns out there's an empty string in the list of arguments to cc. How odd!

jimbxb:
More digging... I wanted to remove the branch, just incase. Here is an even more MWE that has a similar issue...

```
def {noinline} mv(x0:T, ?x1:T, ?b:bool) {
    ?x1 = x0
    ?b = true
}

mv(1.0, ?x:float, ?b)
!println(x)
!println(b)

!nl

def {noinline} mv(x0:T, ?b:bool, ?x1:T) {
    ?x1 = x0
    ?b = true
}

mv(1.0, ?b, ?x)
!println(x)
!println(b)
```

Output:
```
1.000000
false
1.000000
true
```
Ths should be the same, but the first one has issues...

LLVM:
```
define external fastcc void @"type_generics.<0>"() {
  %"tmp#16##0" = tail call fastcc {double, i1} @"type_generics.mv<0>"(double 1.0)
  %"x##0" = extractvalue {double, i1}%"tmp#16##0", 0
  %"b##0" = extractvalue {double, i1}%"tmp#16##0", 1
  call ccc void @print_float(double %"x##0")
  call ccc void @putchar(i8 10)
  tail call fastcc void @"wybe.bool.print<0>"(i1 %"b##0")
  call ccc void @putchar(i8 10)
  call ccc void @putchar(i8 10)
  %"tmp#17##0" = tail call fastcc {i1, double} @"type_generics.mv<1>"(double 1.0)
  %"b##1" = extractvalue {i1, double}%"tmp#17##0", 0
  %"x##1" = extractvalue {i1, double}%"tmp#17##0", 1
  call ccc void @print_float(double %"x##1")
  call ccc void @putchar(i8 10)
  tail call fastcc void @"wybe.bool.print<0>"(i1 %"b##1")
  call ccc void @putchar(i8 10)
  ret void
}

define external fastcc {i64, i1} @"type_generics.mv<0>"(i64 %"x0##0") {
  %"tmp#1##0" = insertvalue {i64, i1} undef, i64 %"x0##0", 0
  %"tmp#2##0" = insertvalue {i64, i1} %"tmp#1##0", i1 1, 1
  ret {i64, i1} %"tmp#2##0"
}

define external fastcc {i1, i64} @"type_generics.mv<1>"(i64 %"x0##0") {
  %"tmp#1##0" = insertvalue {i1, i64} undef, i1 1, 0
  %"tmp#2##0" = insertvalue {i1, i64} %"tmp#1##0", i64 %"x0##0", 1
  ret {i1, i64} %"tmp#2##0"
}

```

So perhaps it seems that generics arent the issue, as the floats are printed out the same. I guess its the unpacking of the boolean thats an issue

jimbxb:
Okay, so it's not just my machine. The same test case fails!

pschachte:
Mine, too (AARM64).  But you fixed the problems with Ubuntu, which is a huge help!  Thank you!

I think you're right that the problem is in converting between float and generic types.  I've fixed that problem in several places, but obviously not everywhere.

I remember discussing type unification with you several years ago, specifically regarding type variables, and I think the problem actually lies there.  We've been doing a lot of patching up of type difficulties in LLVM generation that I think should be handled explicitly in the LPVM code.  I'm thinking now that the right solution is to consider "generic(T)" to be a separate type, distinct from T, that permits automatic conversion during type checking.  So passing a float (or anything else but another generic) where a generic type is expected generates an explicit conversion instruction, and similarly for conversion in the other direction.  But my plan for now is just to hack the LLVM generation to fix this bug and then merge the pull request, and add this as an issue to be addressed later.

jimbxb:
When we generate a call's argument, if the Wybe types aren't the same we should be casting the argument to the expected (callee) type.

When we receive a returned value, we should cast it if the types don't align as well. This should be the case for all of the unpacking as well -- if the member's type differs, cast the unpacked value.

pschachte:
I've been trying to keep the LLVM generation clean and lean.  By that I mean just using the known types of variables and values to determine the generated code.  Unfortunately, we have cases where we generate LPVM such that a variable might have one type when assigned and a different type when used.  That means I've had to track the type of a variable when it is assigned, and use that in place of the type decoration when it is used.  We also have places where a value or variable of one type is passed as an argument to a proc that expects a different type as input or output.  Both of these things happen in cases where one mention of the variable has AnyType as type (meaning it's generic), and another mention has a concrete type.  So I don't think using AnyType as the type of a generic value is a good idea.  I think having a variable or value winding up with AnyType should be considered to be an error.  Instead, I think we need to use a distinct new kind of TypeSpec, something like `Generic TypeSpec`, for generic values.  Then something like append would have arguments of type `Generic AnyType`, which could be unified with, say, `Generic Float` to unify the type variable with `Float`.  At the LPVM level, there would be an explicit conversion between `Generic Float` and `Float`, meaning there would be two different LPVM variables with different types.  That should allow us to type variables consistently, and to ensure that argument types always match the corresponding parameter types, which should allow LLVM generation to be a lot simpler.

BTW, LLVM generation does bitcasting and conversion (eg, sign extending) on every call and return.  Again, I think it would be better to put this in the LPVM, so the LLVM could be pretty simple (there is some inevitable complication in the distinction between single assignment and *static* single assignment).

pschachte:
That's exactly what I'm doing.  But because the type attached to a variable can be different where it is assigned from where it is used (and because llc chucks a wobbly if that happens in the LLVM code), I need to track the assigned types and ignore the attached ones.  The type checker should ensure that all uses of a variable have the same type, but then when the variable is passed to a generic proc, we need to:  (1) bind the generic type variable, so we know when outputs have the same types, and (2) handle the type conversion between the known type and the generic type.  We can't just bind the type variable and then think it's OK to pass that where a generic is expected.

pschachte:
I'll just note that my latest push to the branch fixes both of your examples, but this code still doesn't work:
```
def {noinline} app(f:(I, ?J), i:I):J = f(i)

!println(app({@=?@}, 42.0))
```
It compiles, but prints 0.00000 instead of 42.0000.

jimbxb:
Interesting...

For my implementation of HO, all arguments need to be coerced into ints, as though they were generically typed. This is to allow for a uniform representation with generics, as the same closure proc can be passed as a generic or concretely typed parameter. 

I haven't checked the implementation, but this would be where I would check first.

pschachte:
Yep, I think that's the problem.  The LPVM HO call looks like this:

```
    foo.app<0>(foo.#anon#1<1><>:(wybe.float, ?wybe.float), 42.0:wybe.float, ?tmp#0##0:wybe.float) #0 @foo:3:10
```

and app looks like this:

```
0: foo.app<0>
app(f##0:(I, ?J) <{}; {}; {0}>, i##0:I <{}; {}; {1}>, ?#result##0:J <{}; {}; {0, 1}>)<{}; {}; {}>:
  AliasPairs: []
  InterestingCallProperties: []
    ~f##0:(I, ?J)(~i##0:I, ?#result##0:J) #0 @foo:1:40
```

So it looks like app is being called with a float as input, but the parameter is generic, so it's expecting an i64.

pschachte:
Actually, I see I've dealt with that problem.  Here's the code I'm generating for the call to app:
```
  %"tmp#5##0" = tail call fastcc i64 @"foo.app<0>"(i64 ptrtoint( ptr @"closure#0" to i64 ), i64 bitcast( double 42.0 to i64 ))
```
so I _am_ converting the float to an i64.  The problem is actually the trampoline:
```
proc #anon#1 > {inline} (1 calls)
1: foo.#anon#1<1>
#anon#1(anon#1#1##0:wybe.float, ?anon#1#2##0:wybe.float)<{}; {}; {}>:
  AliasPairs: []
  InterestingCallProperties: []
    foreign llvm move(~anon#1#1##0:wybe.float, ?anon#1#2##0:wybe.float) @foo:3:15
```
which is compiled in the obvious way:
```
define external fastcc double @"foo.#anon#1<1>"(i64 %"#env##0", double %"anon#1#1##0") {
  ret double %"anon#1#1##0"
}
```
Because this is called from HO code, it will actually receive an i64 as input, which it must bitcast to a double, and must finally bitcast its double output to i64 before returning it.  That's not difficult, if there's some way for me to know that it's a trampoline, so its declared parameter types do not actually reflect its input and output types.  Or am I missing something?

jimbxb:
The trampolines should have a flag in their proc def marking them as a "closure" from memory. Each of those parameters should be an i64, and then bitcast accordingly when used inside.

pschachte:
OK, I've found the code that's checking for that and rewriting the trampoline.  It's handled in a function `closeClosure` that I adapted from `Blocks.hs`; I assume you wrote that?  It doesn't actually alter the parameter types, but it's possible I removed code to do that at some point.  I'll put in code to do those transformations on entry and exit.

What about the "free" variables (that are stored in the closure environment)?  Are they stored as their intended types, or as generic (address) values needing the same transformations done?

jimbxb:
From the diffs I'm reading, Blocks has been removed, so would have to be ported. 

I did write that code, a while back now, and it does seem to overwrite the Param types with AnyType, but it never propagates that back into the Compiler's state.

I think that free Params are all smashed into ints as well. This was probably done for simplicity's sake for a uniform representation. You might be able to optimise this into a more compact representation, but I am unsure because the first element of the closure is always a pointer (and hence word sized)

pschachte:
The new `LLVM.hs` module is a rewrite of Blocks.hs and Codegen.hs to directly generate .ll files.  I pinched your `closeClosure` function from Blocks, but I guess I ripped out the part that changed param types to AnyType because it was causing problems.  Wrong fix for the problems in LLVM generation, I guess.

Not storing the modified LPVM back into the module is not a problem, because we're just going to generate LLVM from it and quit.

The code that unpacks the closure assumes that each element is a single full word.  It's extracting each element as its expected type, not as an AnyType, so I think that'll need to be fixed, too.  Bitcasting something smaller than a word to a word to store it, and then reading it back as its intended size, may go wrong for the wrong endianness.

jimbxb:
LGTM!

PR #474, opened 2024-10-21
Read empty and singleton strings specially

pschachte:
These short strings are read using the appropriate
constructor, rather than the general buffer
constructor.

Other changes include optimising foreign lpvm cast
instructions applied to constant chars, integers
and floats, as they get generated in this case.
This avoids several instructions in calling
unboxed constructors with constant arguments.

Also use the -n wybemk flag in final-dump-test.sh,
to turn off colourising, making diffs easier to
read.

There are numerous code improvements in final dump
tests, including allowing specialisation to kick
in in many places, since unboxed values cannot be
aliased.

There's one unfortunate change in output for
constant_type_constraint_error.wybe:  the error
message for `?a = "s":int` now complains about a
type error in the call to wybe.string.singleton,
which doesn't appear on the line.  I think this is
tolerable for now, as it would be difficult to
fix.


pschachte:
Woops, please ignore.

PR #475, opened 2024-10-21
Specially handle short strings in the parser

### LINKED ISSUE #473: Empty and singleton string constants should generate `empty` and `singleton(c)` instead of `buffer(...)`

pschachte:
These short strings are read using the appropriate constructor, rather than the general buffer
constructor.  So the empty and singleton
constructors are now public.

Other changes include optimising foreign lpvm cast instructions applied to constant chars, integers
and floats, as they get generated in this case.
This avoids several instructions in calling
unboxed constructors with constant arguments.

Also use the -n wybemk flag in final-dump-test.sh, to turn off colourising, making diffs easier to
read.

There are numerous code improvements in final dump tests, including allowing specialisation to kick
in in many places, since unboxed values cannot be
aliased.

There's one unfortunate change in output for
constant_type_constraint_error.wybe: the error
message for ?a = "s":int now complains about a
type error in the call to wybe.string.singleton,
which doesn't appear on the line. I think this is
tolerable for now, as it would be difficult to
fix.

Closes #473

### DIFF
--- a/src/BodyBuilder.hs
+++ b/src/BodyBuilder.hs
@@ -19,6 +19,7 @@ import Snippets ( boolType, intType, primMove )
 import Util
 import Config (minimumSwitchCases, wordSize)
 import Options (LogSelection(BodyBuilder))
+import Data.Char ( ord )
 import Data.Map as Map
 import Data.List as List
 import Data.Set as Set
@@ -232,11 +233,6 @@ instance Show Constraint where
         = show v ++ ":" ++ show t ++ " = " ++ show a
     show (NotEqual v t a)
         = show v ++ ":" ++ show t ++ " ~= " ++ show a
-
-
--- negateConstraint :: Constraint -> Constraint
--- negateConstraint (Equal v n)    = NotEqual v n
--- negateConstraint (NotEqual v n) = Equal v n
 
 
 ----------------------------------------------------------------
@@ -954,6 +950,7 @@ updateVariableFlows prim = do
 --  performing the operation at compile-time.
 simplifyForeign ::  String -> ProcName -> [Ident] -> [PrimArg] -> Prim
 simplifyForeign "llvm" op flags args = simplifyOp op flags args
+simplifyForeign "lpvm" op flags args = simplifyLPVM op flags args
 simplifyForeign lang op flags args = PrimForeign lang op flags args
 
 
@@ -1130,6 +1127,8 @@ simplifyOp "fdiv" _ [ArgFloat n1 ty, ArgFloat n2 _, output] =
 simplifyOp "fdiv" _ [arg, ArgFloat 1 _, output] =
   primMove arg output
 -- Float comparisons
+simplifyOp "fcmp_ord" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
+  primMove (boolConstant True) output
 simplifyOp "fcmp_oeq" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
   primMove (boolConstant $ n1==n2) output
 simplifyOp "fcmp_one" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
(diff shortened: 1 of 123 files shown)


src/Parser.hs
```diff
@@ -1219,6 +1219,11 @@ termToExp (Foreign pos lang inst flags args) =
 termToExp (IntConst pos num) = Right $ Placed (IntValue num) pos
 termToExp (FloatConst pos num) = Right $ Placed (FloatValue num) pos
 termToExp (CharConst pos char) = Right $ Placed (CharValue char) pos
+termToExp (StringConst pos "" DoubleQuote)
```
jimbxb, line 1222: Can this be moved to Normalise? I feel like the parser shouldn't have too much knowledge of the library

src/BodyBuilder.hs
```diff
@@ -1130,6 +1127,8 @@ simplifyOp "fdiv" _ [ArgFloat n1 ty, ArgFloat n2 _, output] =
 simplifyOp "fdiv" _ [arg, ArgFloat 1 _, output] =
   primMove arg output
 -- Float comparisons
+simplifyOp "fcmp_ord" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
```
jimbxb, line 1130: Duplicate of below, should be included with the others anyway

src/BodyBuilder.hs
```diff
@@ -1142,9 +1141,39 @@ simplifyOp "fcmp_ogt" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
   primMove (boolConstant $ n1>n2) output
 simplifyOp "fcmp_oge" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
   primMove (boolConstant $ n1>=n2) output
+simplifyOp "fcmp_ord" _ [ArgFloat n1 _, ArgFloat n2 _, output] =
```
jimbxb, line 1144: What does Ord even do?

test-cases/complex/exp/testcase_multi_specz-drone.exp
```diff
@@ -310,7 +309,7 @@ loop(d##0:drone.drone_info, ch##0:wybe.char)<{<<wybe.io.io>>}; {<<wybe.io.io>>};
 
 
         1:
-            wybe.string.print<0>("(":wybe.string)<{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}> #14 @drone:nn:nn
+            wybe.string.print<0>[410bae77d3](1187:wybe.string)<{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}> #14 @drone:nn:nn
```
jimbxb, line 312: Sweet. This always bothered me a little

wybelibs/wybe/string.wybe
```diff
@@ -18,7 +18,7 @@ constructors empty
              # ^ A concatenation of two strings
            | slice(base:_, range:range)
              # ^ A slice of a string under the given range
-           | singleton(c:char)
+           | pub singleton(c:char)
```
jimbxb, line 21: Slightly unfortunate we can't have the two pubs together. I want empty (really, all const ctors) first, but buffer needs to be the first non-const.

jimbxb (changes requested):

src/Parser.hs
jimbxb: Moving this after type checking (maybe in Clause?) would solve the type error issue, too. But that doesn't make as much sense as in Normalise.

wybelibs/wybe/string.wybe
jimbxb: I think a comment why these need to be public is good too. 

I realise the reason why buffer needs to be the first non-const is not clear here, so some elucidation on that would be helpful to pick up now.

src/BodyBuilder.hs
pschachte: fcmp_ord tests that neither value is a NaN.  I've updated WYBE.md to clarify.

src/Parser.hs
pschachte: OK, I'll move it to Expansion, after type checking but early enough for the inlining to still work.

jimbxb:
Sorry for being a bit late to the party, but I'm not sure Expansion is the right place to do this. I believe this string specialisation pass should only ever happen once, as the transformation is idempotent. Expansion may only get ran once at the moment (correct me if I am wrong), but theoretically it could get ran multiple times, so the specialisation pass would be redundant after the first Expansion pass in the module. I realise that we haven't been focusing on making the compiler the most efficient, but I feel we should avoid inefficiencies if we are aware of them.

pschachte:
Expansion is only run once (not positive there are no circumstances when it is repeated, but I think it's not).  But you're right, we may want to repeat it.  This shouldn't be a problem, though, since after this is done, we don't have the ArgString constructor for empty or singleton strings, so the change is idempotent.  We're already scanning every argument during expansion, to update variables, so looking at ArgStrings and checking them for empty or singleton lists is a minimal added overhead.  And if we moved it somewhere else, it would impose a similar overhead.

PR #476, opened 2024-12-12
Const structs

pschachte:
Generate LLVM constant structures in place of code to create structures all of whose contents are constants.  This also applies to constant closures, so the code to optimise them in LLVM generation is now not needed, and to strings of length >1.


src/AST.hs
```diff
@@ -3406,14 +3635,49 @@ argGlobalFlow varFlows (ArgClosure pspec args _) = do
...
+    params <- getPrimParams pspec
+    let nArgs = length fields
+        (closedParams, freeParams) = List.splitAt nArgs params
+    if any (\(PrimParam _ ty flow _ _) ->
+            isInputFlow flow && isResourcefulHigherOrder ty) freeParams
+    then return univGlobalFlows
+    else do
+        gFlows <- getProcGlobalFlows pspec
+        globalFlowsUnions . (gFlows:) <$> mapM constGlobalFlows fields
+constsGlobalFlows fields = return emptyGlobalFlows -- XXX need to recurse!
```
jimbxb, line 3668: Is it also possible there are global flows in the other fields? I think it's possible that there is another function pointer nested inside one of the other fields, and I think that should be considered.

test-cases/complex/exp/testcase_multi_specz-drone.exp
```diff
@@ -13,7 +13,7 @@
 ----------------------------------------------------------------------
 (0, 0, 0) #0
 (-1, -6, 3) #100
-** malloc count: 1
+** malloc count: 100
```
jimbxb, line 16: This is unfortunate.

jimbxb (commented):
I will get back to the large files that I haven't read in some time. I also need to go through the changes in at least one of the larger exp files in more detail, but they seems sensible!

It's most unfortunate that there's a 100x in the mallocs in the drone test case. I suppose that's harder to get back now that the first drone is static and the compiler isn't smart enough to use a specialisation after that fact :/

test-cases/complex/exp/testcase_multi_specz-drone.exp
pschachte: Very much so.  The two possible fixes I can think of would be to get multiple specialisation working for the recursive call, or to turn the constant back into code to create a fresh structure when that would allow the update to be destructive.  Both of these seem pretty difficult to wrangle, so I put in the PR without fixing this.

src/LLVM.hs
```diff
@@ -271,47 +277,75 @@ preScanProcs = do
...
+argConstValue ArgVar{} = return Nothing
+argConstValue (ArgInt n ty) = do
+    sz <- typeRepSize <$> lift (typeRepresentation ty)
+    return $ Just $ IntStructMember n (sz `div` 8)
+argConstValue (ArgFloat n ty) = do
+    sz <- typeRepSize <$> lift (typeRepresentation ty)
+    return $ Just $ FloatStructMember n (sz `div` 8)
+argConstValue ArgClosure{} = return Nothing -- const closures already handled
+argConstValue (ArgGlobal info _) = do
+    -- XXX is ArgGlobal a constant?  Does it give the address, or value, of the
```
jimbxb, line 340: I think this has come up before. I can't recall exactly what I thought last time, but I think that the address makes sense -- in order to get the value you must `load`.

src/AST.hs
```diff
@@ -1417,11 +1454,96 @@ emptyModule = Module
...
+
+instance Show StructID where
+    show structID@(StructID _ _ (Just exp))= show exp
+    show structID = structConstName structID
+
+
+-- | The LLVM constant name for the specified StructID
+structConstName :: StructID -> Ident
+structConstName (StructID md num _) =
+    specialName2 (showModSpec md) $ "constant" ++ specialChar : show num
```
jimbxb, line 1476: We need a `specialName{N}` variant, for N > 2. Seems annoying...

Maybe

```
specialNames :: [String] -> String
specialNames = intercalate [specialChar]
```
is more ergonomic.

src/LLVM.hs
```diff
@@ -428,55 +462,53 @@ writeAssemblyPrologue = do
...
-    trustFromJust ("lookupConstant " ++ show spec) <$> tryLookupConstant spec
-
-
--- | Find the global constant that holds the value of the specified string
--- constant.
-tryLookupConstant :: StaticConstSpec -> LLVM (Maybe Ident)
-tryLookupConstant spec =
-    gets $ Map.lookup spec . constNames
+writeConstDeclaration :: StructID -> LLVM ()
+-- writeConstDeclaration spec@(WybeStringSpec str) n = do
```
jimbxb, line 475: Delete?

src/LLVM.hs
```diff
@@ -785,7 +817,7 @@ writeWybeCall wybeProc args pos = do
 -- | Generate a Wybe proc call instruction, or defer it if necessary.
 writeHOCall :: PrimArg -> [PrimArg] -> OptPos -> LLVM ()
 writeHOCall (ArgClosure pspec closed _) args pos = do
-    -- NB:  this case doesn't seem to ever occur
+    -- NB:  this case doesn't seem to ever occur -- probably handled earlier
```
jimbxb, line 820: That's right. The body builder should always first-order-ify these cases. I think I added it a while back, probably before I tackled it in the body builder! Should be safe to remove.

src/LLVM.hs
```diff
@@ -1495,9 +1546,7 @@ neededFreeArgs pspec args = lift $ do
 
 -- | Return the representation for the specified type
 typeRep :: TypeSpec -> LLVM TypeRepresentation
-typeRep ty =
-    trustFromJust ("lookupTypeRepresentation of unknown type " ++ show ty)
-      <$> lift (lookupTypeRepresentation ty)
+typeRep = lift . typeRepresentation
```
jimbxb, line 1549: Do we really need this shorthand now?

src/LLVM.hs
```diff
@@ -1775,10 +1825,10 @@ type ConstSpec = (PrimArg,TypeRepresentation)
 -- | The LLVM State monad
 data LLVMState = LLVMState {
         -- These values apply to a whole module (and submodules)
-        allConsts :: Set StaticConstSpec,
+        allConsts :: Set StructID,
                                      -- ^ Static constants appearing in module
-        constNames :: Map StaticConstSpec Ident,
-                                    -- ^ local name given to static constants
+        -- constNames :: Map StaticConstSpec Ident,
```
jimbxb, line 1830: Delete?

jimbxb (changes requested):

test-cases/execution/nbody.exp
```diff
@@ -2,4 +2,4 @@ n: energy: -0.169075
 energy: -0.169088
 
 malloc count: (should be 5005 with multi-specz and 10006 without it)
-5005
```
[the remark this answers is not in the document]
jimbxb: Can we update the comment above?

src/AST.hs
```diff
@@ -3345,16 +3482,108 @@ data PrimArg
...
+constValueSize (PointerStructMember _) = wordSizeBytes
+constValueSize (FnPointerStructMember _) = wordSizeBytes
+constValueSize (UndefStructMember size) = size
+constValueSize (GenericStructMember _) = wordSizeBytes
+
+
+-- | Return the representation of a StructMember
+constValueRepresentation :: ConstValue -> TypeRepresentation
+constValueRepresentation (IntStructMember _ size) = Bits $ size * 8
+constValueRepresentation (FloatStructMember _ size) = Floating $ size * 8
```
jimbxb, line 3546: I recall having constants for these `8`s. Can we use them here too, for clarity?

src/AST.hs
```diff
@@ -3761,6 +4020,18 @@ envPrimParam = PrimParam envParamName AnyType FlowIn Ordinary (ParamInfo False e
 makeGlobalResourceName :: ResourceSpec -> String
 makeGlobalResourceName spec = specialName2 "resource" $ show spec
 
+-- These are defined in Snippets.hs, but we can't import that as it imports AST
```
jimbxb, line 4023: Non-recursive modules is painful 😖

src/BodyBuilder.hs
```diff
@@ -838,30 +1030,116 @@ addInstrToState ins st@BodyState{buildState=bld@Forked{bodies=bods}} =
...
+-- | Try to guess the original expression that led to the specified constant
+-- structure members.  Note that members are in reverse order.  This code relies
+-- on the current implementations of some library types and functions.
+guessExp :: [(Int, ConstValue)] -> BodyBuilder (Maybe Exp)
+guessExp members = do
+    logBuild $ "Guessing expression for structure members " ++ show members
+    case members of
+      [(n,PointerStructMember id),(0,IntStructMember len _)] ->
+        lift (lookupConstInfo id) >>= \case
+          Just (CStringInfo str) | len == fromIntegral (length str) -> do
```
jimbxb, line 1094: Is there ever a case where this isn't true?

I'm overall not sure why we need this...

jimbxb:
I'm reviewing this via my phone at the airport, so can't access my laptop. Unfortunately, it seems that after a certain number of files, GitHub refuses to load diffs. I noticed, by chance in one of the commits a change that is odd. This should be the reference to it:

https://github.com/pschachte/wybe/blob/const_structs/test-cases%2Ffinal-dump%2Fearly_error.exp#L74

Any chance we can avoid these changes? I assume the issue is before types, so perhaps in Flatten/Normalise, where string constants are replaced with tmp variables (that hold the constant value). I think this is also causing one of the error messages changing too, that I commented on before.

src/Flatten.hs
```diff
@@ -505,8 +506,14 @@ flattenExp expr@(IntValue _) ty castFrom pos =
     return $ typeAndPlace expr ty castFrom pos
 flattenExp expr@(FloatValue _) ty castFrom pos =
     return $ typeAndPlace expr ty castFrom pos
-flattenExp expr@(StringValue _ _) ty castFrom pos =
+flattenExp expr@(ConstStruct _) ty castFrom pos =
     return $ typeAndPlace expr ty castFrom pos
+flattenExp expr@(StringValue str WybeString) ty castFrom pos = do
+    exp' <- lift $ stringExpr str
```
jimbxb, line 512: I suppose this is what is causing the errors around string constants, as extra "unreachable" statements are introduced. 

I don't know how we can avoid this, unfortunately. Perhaps we can delay this until Clause? That has implications for the const structures that are referenced in the module, I believe.

src/AST.hs
pschachte: But specialName puts the special character at the front, while specialName2 puts it between the names.  Either way, only one special character is inserted.

src/AST.hs
```diff
@@ -1417,11 +1454,96 @@ emptyModule = Module
...
+    return $ Typed (Fncall ["wybe","string"] "empty" False [])
+                stringType Nothing
+stringExpr [ch] = do
+    let arg = Unplaced $ CharValue ch
+    return $ Typed (Fncall ["wybe","string"] "singleton" False [arg])
+                stringType Nothing
+stringExpr str = do
+    cStringArg <- Unplaced <$> cStringExpr str
+    return $ Typed
+            (Fncall ["wybe","string"] "unsafe_cstring_to_string" False
```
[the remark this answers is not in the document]
pschachte: Yeah, I wanted to put "unsafe" in there, since this bypasses the check for length >= 2.  Not that anything terrible would happen now if we used a buffer for a 1 or 0 character string, but it might be useful to be able to count on 0 or 1 characters using the specialised representation.

src/AST.hs
pschachte: I can't find any definition of a constant for the number of bits per byte, but I've defined one now.

src/AST.hs
pschachte: Yup, really annoying sometimes.

src/LLVM.hs
pschachte: We don't need it, but it's used enough to seem worthwhile.

src/Flatten.hs
pschachte: That's right.  Unfortunately, Clause can't handle non-atomic expressions, so it can't turn a string constant into a proc call.  Fixing that would entail merging flattening into LPVM generation, and rewriting the type checker and unbrancher to operate on unflattened code.  That's too big a change.

jimbxb: Quite unfortunate. I was hoping that clause could at least handle the String constant expressions. I guess I have forgotten how clause works...

src/Clause.hs
```diff
@@ -310,8 +311,9 @@ compileArg exp pos = shouldnt $ "Compiling untyped argument " ++ show exp
 compileArg' :: TypeSpec -> Exp -> OptPos -> ClauseComp [PrimArg]
 compileArg' typ (IntValue int) _ = return [ArgInt int typ]
 compileArg' typ (FloatValue float) _ = return [ArgFloat float typ]
-compileArg' typ (StringValue string v) _ = return [ArgString string v typ]
```
jimbxb, line 313: Ah, I see. So the `compileArg'` function has no ability to "emit" an extra prim.

Can't we modify it to allow emission of both a `([PrimArg], [Prim])`, then prepend all the collected `[Prim]` from all args before the resultant `Prim` which uses this Arg? That way, we can omit an extra call to whichever string constructor we required

pschachte: That is currently handled by flattening and mode checking (remember, expressions can be outputs, in which case the flattened statement must come after the current statement), so this would be a major architectural change.  I'm working on a better fix, which involves annotating expressions with their untransformed version, so it would fix many other places where we confusingly show transformed code in error messages.  I think the right fix would be to incorporate the source code into the OptPos type, but that would be a major change to the parser.  We'll see what I can come up with.

src/AST.hs
```diff
@@ -37,8 +37,8 @@ module AST (
   setExpFlowType,
```
pschachte, line 37:
```suggestion
  setExpFlowType, sourceAnnotateExp,
```

src/AST.hs
```diff
@@ -47,7 +47,12 @@ module AST (
   OptPos, Placed(..), place, betterPlace, content, maybePlace, rePlace, unPlace,
```
pschachte, line 47:
```suggestion
  SourceInfo(..), OptPos, Placed(..), place, betterPlace, content,
  maybePlace, rePlace, unPlace,
```

src/AST.hs
```diff
@@ -161,9 +191,9 @@ import           System.Console.ANSI
 
 import           GHC.Generics (Generic)
 
--- import qualified LLVM.AST as LLVMAST
 import Data.Binary (Binary)
```
pschachte, line 194:
```suggestion
import Data.Binary (Binary)
import GHC.Profiling (startProfTimer)


-- |The tokens of the wybe language, each carrying its source position.
data Token = TokFloat Double SourcePos          -- ^A floating point number
              | TokInt Integer SourcePos        -- ^An integer
              | TokString StringDelim String SourcePos
                                                -- ^A string with its delimiter
              | TokChar Char SourcePos          -- ^A character constant
              | TokIdent String SourcePos       -- ^An identifier
              | TokLBracket BracketStyle SourcePos
                                                -- ^Some kind of left bracket
              | TokRBracket BracketStyle SourcePos
                                                -- ^Some kind of right bracket
              | TokComma SourcePos              -- ^A comma
              | TokPeriod SourcePos             -- ^A period (full stop)
              | TokSymbol String SourcePos      -- ^A symbol made up of
                                                --  non-identifier chars
              | TokSeparator Bool SourcePos     -- ^A statement separator
                                                -- indicating whether explicit
              | TokError String SourcePos       -- ^A lexical error
    deriving (Eq, Ord, Generic)

-- |The different string delimiters.
data StringDelim = DoubleQuote
                 | BackQuote
                 | LongQuote String
                 | IdentQuote String StringDelim
               deriving (Eq, Ord, Generic)


-- |The different kinds of brackets.
data BracketStyle = Paren | Bracket | Brace
                  deriving (Eq, Ord, Generic)

```

src/AST.hs
```diff
@@ -1072,28 +1101,27 @@ isSpecialResource :: ResourceSpec -> Bool
 isSpecialResource res = res `Set.member` specialResourcesSet
 
 
-callFileName :: Placed Stmt -> Exp
+callFileName :: Placed Stmt -> Compiler Exp
 callFileName pstmt =
-    (`StringValue` CString)
-    $ maybe "Unknown file" (takeBaseName . sourceName) (place pstmt)
+    cStringExpr $ maybe "Unknown file" (takeBaseName . sourceName) (place pstmt)
```
pschachte, line 1123:
```suggestion
    cStringExpr $ maybe "**UNKNOWN FILE**" (takeBaseName . sourceName) (realPlace pstmt)
```

src/AST.hs
```diff
@@ -1072,28 +1101,27 @@ isSpecialResource :: ResourceSpec -> Bool
...
 
-callFileFullName :: Placed Stmt -> Exp
+callFileFullName :: Placed Stmt -> Compiler Exp
 callFileFullName pstmt =
-    (`StringValue` CString)
-    $ maybe "Unknown file" sourceName (place pstmt)
+    cStringExpr $ maybe "Unknown file" sourceName (place pstmt)
```
pschachte, line 1127:
```suggestion
    cStringExpr $ maybe "**UNKNOWN FILE**" sourceName (realPlace pstmt)
```

src/AST.hs
```diff
@@ -1072,28 +1101,27 @@ isSpecialResource :: ResourceSpec -> Bool
...
 
-callLineNumber :: Placed Stmt -> Exp
+callLineNumber :: Placed Stmt -> Compiler Exp
 callLineNumber pstmt =
-    IntValue $ fromIntegral $ maybe 0 sourceLine (place pstmt)
+    return $ Typed (IntValue $ fromIntegral $ maybe 0 sourceLine (place pstmt))
```
pschachte, line 1131:
```suggestion
    return $ Typed (IntValue $ fromIntegral $ maybe 0 sourceLine (realPlace pstmt))
```

src/AST.hs
```diff
@@ -1072,28 +1101,27 @@ isSpecialResource :: ResourceSpec -> Bool
...
+                intType Nothing
 
-callColumnNumber :: Placed Stmt -> Exp
+callColumnNumber :: Placed Stmt -> Compiler Exp
 callColumnNumber pstmt =
-    IntValue $ fromIntegral $ maybe 0 sourceColumn (place pstmt)
+    return $ Typed (IntValue $ fromIntegral $ maybe 0 sourceColumn (place pstmt))
```
pschachte, line 1136:
```suggestion
    return $ Typed (IntValue $ fromIntegral $ maybe 0 sourceColumn (realPlace pstmt))
```

src/AST.hs
```diff
@@ -1072,28 +1101,27 @@ isSpecialResource :: ResourceSpec -> Bool
...
+                intType Nothing
 
-callSourceLocation :: Bool -> Placed Stmt -> Exp
+callSourceLocation :: Bool -> Placed Stmt -> Compiler Exp
 callSourceLocation full pstmt =
-    (`StringValue` CString)
-    $ maybe "unknown location" (showSourcePos full) (place pstmt)
+    cStringExpr $ maybe "unknown location" (showSourcePos full) (place pstmt)
```
pschachte, line 1141:
```suggestion
    cStringExpr $ maybe "unknown location" (showSourcePos full) (realPlace pstmt)
```

src/AST.hs
```diff
@@ -3122,6 +3257,7 @@ data Exp
                -- these two must be the same.
       | Global GlobalInfo
```
pschachte, line 3314:
```suggestion
      | Global GlobalInfo
      | TransformedFrom Exp Exp -- ^A transformed Exp, together with the
                                -- original version, for error messages.
```

src/AST.hs
```diff
@@ -3612,20 +3870,21 @@ stmtsInputs = foldStmts (const . const)
...
 expOutputs FailExpr = Set.empty
 expOutputs (AnonParamVar mbNum flow) =
     if flowsOut flow then Set.singleton ("@" ++ maybe "" show mbNum) else Set.empty
-expOutputs (Global _) = Set.empty
+expOutputs Global{} = Set.empty
 expOutputs AnonProc{} = Set.empty
 expOutputs AnonFunc{} = Set.empty
-expOutputs (Closure _ _) = Set.empty
+expOutputs Closure{} = Set.empty
 expOutputs (Typed expr _ _) = expOutputs expr
```
pschachte, line 3975:
```suggestion
expOutputs (Typed expr _ _) = expOutputs expr
expOutputs (TransformedFrom expr _) = expOutputs expr
```

src/AST.hs
```diff
@@ -3644,20 +3903,21 @@ pexpListOutputs = List.foldr (Set.union . expOutputs . content) Set.empty
...
 expInputs FailExpr = Set.empty
 expInputs (AnonParamVar mbNum flow) =
     if flowsIn flow then Set.singleton (show mbNum) else Set.empty
-expInputs (Global _) = Set.empty
+expInputs Global{} = Set.empty
 expInputs AnonProc{} = Set.empty
 expInputs AnonFunc{} = Set.empty
-expInputs (Closure _ _) = Set.empty
+expInputs Closure{} = Set.empty
 expInputs (Typed expr _ _) = expInputs expr
```
pschachte, line 4009:
```suggestion
expInputs (Typed expr _ _) = expInputs expr
expInputs (TransformedFrom expr _) = expInputs expr
```

src/Flatten.hs
```diff
@@ -505,8 +506,14 @@ flattenExp expr@(IntValue _) ty castFrom pos =
...
+    flattenExp exp' ty castFrom pos
```
pschachte:
```suggestion
    exp'  <- lift $ stringExpr str
    exp'' <- flattenExp exp' ty castFrom pos
    return $ sourceAnnotateExp exp'' expr
```

src/Flatten.hs
```diff
@@ -639,10 +646,20 @@ flattenExp (ForeignFn lang name flags exps) ty castFrom pos = do
     flattenCall (ForeignCall lang name flags) True ty castFrom pos exps
 flattenExp (Typed exp AnyType _) ty castFrom pos = do
     flattenExp exp ty castFrom pos
```
pschachte, line 655:
```suggestion
    flattenExp exp ty castFrom pos
flattenExp (TransformedFrom exp src) ty castFrom pos = do
    exp' <- flattenExp exp ty castFrom pos
    return $ maybePlace (sourceAnnotateExp exp' src) pos
```

src/AST.hs
pschachte: Can you explain what you mean?  I confess I don't fully understand global flows and where they might be hiding.

src/BodyBuilder.hs
pschachte: It's been so long I can't remember, but from the fn doc it looks like I was leaving the door open to handle other library types, or for us to change the string type.

test-cases/final-dump/bug228-type.exp
```diff
@@ -1,4 +1,4 @@
 Error detected during type checking of module(s) bug228-type
 final-dump/bug228-type.wybe:5:2: Type error in call to proc =, argument 1
 final-dump/bug228-type.wybe:5:2: Type error in call to proc =, argument 2
-final-dump/bug228-type.wybe:5:2: Type of "1.0" incompatible with ?bar
+final-dump/bug228-type.wybe:5:2: Type of tmp#0:wybe.string incompatible with ?bar
```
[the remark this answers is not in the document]
pschachte: Yes, it is, but it's hard to fix because flattening loses the original code.  This causes lots of poor error messages.  But I'd like to tolerate this for now and fix it in a later pull.

jimbxb (approved):
I took the liberty of merging in master. All still seems to work!

pschachte (commented):
LGTM

jimbxb:
Seems a bunch of comments were just added. If I find some time to go through them I shall, but it would be great if you could see if they're still needed @pschachte

pschachte:
Yeah, sorry, I'm trying to wrap up this PR, and have gotten stuck trying to deal with all the clashes merging master back into this branch.  It's painful when a branch falls far behind.

jimbxb:
I suspect there's a missing prescan step in the LLVM generation, where nested constants aren't being added as globals. There's probably a recursion on a primarg that is missed.

PR #479, opened 2025-04-20
Add --llc-path and --log-file options

### LINKED ISSUE #477: Add compiler flag to control path of llc binary
Currently, `llc` is invoked using a starting string, and is resolved using the `$PATH`. It would be nice if we can support a compiler flag to modify this location, `--llc-path={String}`. This would be useful if multiple `llc` versions are installed, for instance. If unspecified, the behaviour should be as if it were `"llc"`

### LINKED ISSUE #478: Route logs to a specified file.
Currently, logs go to `stderr`. This can be rerouted fairly trivially using your favourite shell. It would be convenient to support writing this to a file, which could be specified with a compiler flag, `--log-file={String}`. If specified, the colour-outout mode should be disabled by default, unless manually turned on.

I am not sure on any performance considerations, but I think we can cache a file handle in the compiler's config object, which can default to the file handle for stderr if no log file is specified. This handle can then be used by the `log*` functions used extensively throughout the compiler, and likely only needs to be specified in the generic log function (I forget the name). I think we would want the file should be re-written upon each invocation,  instead of being appended to. 

As a simple extension to this, if the file name is unspecified, a file could be created in `/var/log` (`/var/log/wybe-$date`) and have logs written there instead. The name of this file should be written to stdout/err, and perhaps referenced some way in the generated llvm file for future reference?

jimbxb:
Closes #477 
Closes #478

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -483,8 +483,9 @@ data CompilerState = Compiler {
   errorState :: Bool,            -- ^whether or not we've seen any errors
   modules :: Map ModSpec Module, -- ^all known modules except what we're loading
   underCompilation :: [Module],  -- ^the modules in the process of being compiled
-  unchangedMods :: Set ModSpec   -- ^record mods that are loaded from object
+  unchangedMods :: Set ModSpec,  -- ^record mods that are loaded from object
                                  --  and unchanged.
+  logHandle :: Handle            -- ^handle to write logs to
 }
 
 -- |The compiler monad is a state transformer monad carrying the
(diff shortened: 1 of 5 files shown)


pschachte (approved):
This looks fine, so I'll pull it, despite the problem with the CI system you reported.

pschachte:
BTW, I don't think logging to /var/log is a good idea, given the intent of /var/log and the purpose of wybemk logging.  The former is intended for logging aspects of the normal execution of system programs, while the latter is intended for debugging the compiler.  I don't know of any other compilers that log to /var/log.

pschachte:
Yep.  I always write the log to a file.  You've probably seen me accidentally commit a file called `LOG` to the repo, and then have to delete it later.

PR #482, opened 2025-04-24
(Attempt to) fix CI workflow after github change

pschachte:


jimbxb:
I was hoping that a version bump would be enough to fix CI. Unfortunately, my train ride was only so long and I didn't get a chance to try

PR #484, opened 2025-08-20
Fix #483

### LINKED ISSUE #483: Compiler causes error when multiple specialisation requires library files to be recompiled
For example, when using the installed compiler to compile the execution/string_interpolation test, the compiler crashes (unless run with sudo).

pschachte:
Check writeability of destination object file before writing .ll file, and change destination to cache directory if it's not writeable.


jimbxb (approved):
LGTM. I assume this works on your machine too, which gives us coverage of Ubuntu and MacOS :)

PR #501, opened 2025-11-10
Add efficient array type

pschachte:
This PR renames array.wybe to c_array.wybe, to emphasise that it's intended for accessing arrays that come from C.  Unfortunately, due to limitations in Wybe's uniqueness analysis, mutating c_arrays always requires the data to be copied, even if the c_array is not used again.  The final-dump/new_array_test test case illustrates:  it allocates a million element array, fills it with the number from 0 to 999,999, and sums the contents.  Using c_arrays, this runs for 468 seconds; using the new array implementation, it takes 10 milliseconds.

The new array implementation is a single chunk of memory, with the array size stored in the first word of the memory, followed by the data.  I decided to call this type `array` and rename the original type because Wybe code that creates its own arrays probably wants the faster implementation.

The new implementation is not quite backward compatible.  In particular, you can't directly iterate over the new array type.  I've added a function `values` that returns an iterator that can be iterated over.  It works much like the original array type.  So you need to write `for ?x in arry^values ...` instead of `for ?x in arry ...` to iterate over `arry`.  Otherwise `array` is mostly backward compatible.  I've also written an `array` constructor that copies a c_array to create an `array`, but I haven't tested it yet.

### DIFF
--- a/wybelibs/wybe/array.wybe
+++ b/wybelibs/wybe/array.wybe
@@ -44,7 +44,13 @@ pub def array(x:T, len:int):_(T) = ar where {
     }
 }
 
-# Construct a Wybe array from an c_array.  This will need to copy the data.
+# Construct an array filled with 0s.  Takes advangage of the fact that GC_malloc
+# clears the allocated memory.
+pub def array(len:int):_(T) = ar where {
+    allocate_array(len, ?ar, ?size)
+}
+
+# Construct a Wybe array from a c_array.  This will need to copy the data.
 pub def array(data:c_array(T)):_(T) = ar where {
     allocate_array(data^length, ?ar, ?size)
     ?i = 0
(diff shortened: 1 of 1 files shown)


wybelibs/wybe/array.wybe
```diff
@@ -1,72 +1,150 @@
...
+# Return a sub-array from index `start` (inclusive) to `end` (exclusive).
+# We can't just return a pointer into the original array, because the first word
+# of an array is its length.  So we "pretend" the array actually starts one word
+# before the specified start index, and its length is `end - start`, and use the
+# LPVM mutate operation to store the size where it needs to be.  Wybe's
+# uniqueness analysis will turn that into a destructive update if the original
+# array is unused after the slice operation, otherwise it will copy as much data
+# as is needed for the slice, and then write the size where it belongs.
+pub def {partial} slice(a:_(T), start:int, end:int):_(T) = subarray where {
+    (0 <= start)
```
jimbxb: Can we clip the start to be `max(0, start)`? Similarly for the other bounds. 

I would rather this not fail, and instead just yield an empty result. It means the caller need not ensure the indexes are in bounds.

Perhaps we add a variant of this that does the checking, and move these checks there?

jimbxb (approved):

PR #504, opened 2025-11-19
add WYBEARGS env var, for passing cmdline variables to compiler

### LINKED ISSUE #481: Support a shell variable and/or a file holding compiler switches to automatically apply
These would be applied first, with the actual command-line arguments coming after.  There are now enough compiler switches that it becomes burdensome to type them every time the compiler is run.

There should also be a new switch that says to ignore the contents of the shell variable/file.

jimbxb:
Closes #481


jimbxb:
WY🐻GS

PR #506, opened 2025-11-21
update install guide on Ubuntu for LLVM 18

jimbxb:
I assume a similar patch could be done for the install guide with MacOS, too, but I don't have a machine to test that on!


pschachte (approved):
All good.  I didn't realise it had gotten so out of date.

PR #507, opened 2025-11-23
Move interning of WybeStrings to avoid extranous unreachable warnings

jimbxb:


test-cases/final-dump/resource_error.exp
```diff
@@ -1,2 +1,2 @@
 Error detected during resource checking of module(s) resource_error
-final-dump/resource_error.wybe:2:5: Call to resourceful proc without ! resource marker: wybe.string.<0>println(tmp#0:wybe.string)
+final-dump/resource_error.wybe:2:5: Call to resourceful proc without ! resource marker: wybe.string.<0>println(tmp#0:wybe.string @resource_error:nn:nn)
```
jimbxb: We really shouldnt be showing the whole proc, maybe just the name?

PR #511, opened 2025-11-25
fix(#510): emit LLVM cast for ret

### LINKED ISSUE #510: LLVM variables arent cast when `ret`, if assigned to different type
Test case:
```
use logging

?p = 3
!println(proc(p))
def {noinline, test} proc(p:int) {
    ?f = { !logmsg("$p"); !lognl; (p = @) = ?@ }
    f(2, ?a)
    a
}
```
(no clue if minimal)
yields:
```
llc: error: llc: /home/james/wybe/test-cases/final-dump/bug_llvm_retval.ll:38:10: error: '%tmp#7##0' defined with type 'i64' but expected 'i1'
  ret i1 %"tmp#7##0"
         ^

llc: error: llc: /home/james/wybe/test-cases/final-dump/bug_llvm_retval.ll:38:10: error: '%tmp#7##0' defined with type 'i64' but expected 'i1'
  ret i1 %"tmp#7##0"
         ^
```

jimbxb:
Closes #510 

Still got it 😄


pschachte (approved):
Found and fixed in short order!  Thanks.

PR #521, opened 2025-12-05
Add foreign aliases and foreign function defs

### LINKED ISSUE #467: Allow for foreign proc def to alias the underlying foreign proc
Following from #366.

Currently in the stdlib, most of the Wybe procs that simply call a foreign proc could use the `def foreign lang` syntax. However, as the proc they call has a different name, the syntax sugar can't be applied.

As a proposal, an extended syntax sugar could be like the following:

```
def foreign lang foo=bar(args)
```
This would compile as
```
def foo(args) {
    foreign lang bar(args)
}
```

Of course, proc modifiers and the resource `use` clause should be optional.

### LINKED ISSUE #469: Add function version of foreign function declaration
As a convenience/comfort for users accustomed to thinking of foreign procs as functions, allow a syntax like

    `def foreign` *language* *function*`(`*param*, *param*, ...`)` `:` *type*

as equivalent to

    `def foreign` *language* *function*`(`*param*, *param*, ..., `result`:*type*`)`

If #468 is implemented, this should be extended in the obvious way.

jimbxb:
Closes #467
Closes #469

### DIFF
--- a/test-cases/complex/exp/testcase_multi_specz-drone.exp
+++ b/test-cases/complex/exp/testcase_multi_specz-drone.exp
@@ -149,7 +149,7 @@ module top-level code > public {semipure} (0 calls)
   MultiSpeczDepInfo: [(4,(drone.loop<0>,fromList [NonAliasedParamCond 0 []]))]
     drone.drone_init<0>(?tmp#0##0:drone.drone_info) #0 @drone:nn:nn
     foreign lpvm load(<<wybe.io.io>>:wybe.phantom, ?%tmp#3##0:wybe.phantom) @drone:nn:nn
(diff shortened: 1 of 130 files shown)


WYBE.md
```diff
@@ -2136,24 +2136,40 @@ which is equivalent to defining the following:
...
 
-Note that all parameters must be typed.
+Note that all parameters must be typed. 
 
-Resources can optionally be specified with a `use` clause, that follows the same syntax
-for a regular Wybe procedure definition. Resources are added as additonal arguments to 
-the foreign call after the other arguments, in the order specified in the source code.
-If resources are repeated in the `use` clause, they are added as multiple arguments in 
-the generated foreign call.
+The Wybe function can also use a different name to the foreign procedure, using the following syntax:
```
pschachte: Can you move this after the part about foreign function syntax, and clarify that aliases can be specified for both the procedural and functional syntax?

pschachte (changes requested):

PR #524, opened 2025-12-08
fix(#523): use call inlining when generating a prim

### LINKED ISSUE #523: Procs are tagged with the incorrect purity during clause generation
I noticed this when logging booleans, which is importantly the only `logmsg` overload that doesnt get inlined

Heres an example
```
use logging

def foo(y:bool) use !io {
    !logmsg(y)
}

!foo(true)
```
This should print `true`, but prints nothing.

The issue is that the `logmsg` call when converted into LPVM takes the context's impurity (pure), not the call's (impure). We do calculate it, its just not used! Kinda funny this has made it at least 3 years.

jimbxb:
Closes #523


pschachte (commented):
That test case looks a little suss to me.  It `use`s io to keep it from being eliminated, but doesn't ever actually do any io.  My first reaction is that we should issue a warning for a proc that it `use`s a resource it does not actually use.  And maybe we should do the same for taking an unused parameter.  But maybe I'm being too picky:  an interface may specify parameters or resources that the programmer thinks they may eventually need, and want callers to respect.  We could have some kind of "I know I'm not using this, don't warn me about it" annotation or something to allow this.  What do you think?

For now, can you add some io statement to the body without messing up the test case?

pschachte:
The kind of thing I had in mind for unused resource -- where the resource will or might be necessary eventually, but there's a quick and dirty implementation for now that doesn't use it.  That could also apply to unused parameters.

I like the idea of having a way to suppress each possible instance of each possible warning, rather than suppressing all warning or all instances of a particular kind of warning, on the argument that it makes good documentation to say "I know this isn't great, but I mean to do it".  We could use the Prolog convention for parameters:  if it begins with `_`, we don't warn if it's unused.  That won't work for resources, though.

test-cases/final-dump/bug_523.exp
```diff
@@ -0,0 +1,62 @@
...
+  InterestingCallProperties: []
+    bug_523.impure<0> #0 @bug_523:nn:nn
+
+  LLVM code       :
+
+;; FILE GENERATED BY wybemk 0.2 -- see https://github.com/pschachte/wybe
+; ModuleID = 'bug_523'
+
+source_filename = "!ROOT!/final-dump/bug_523.wybe"
+target triple    ????
```
pschachte: Hmmm, that's wrong.  Can you run `clang --version` for me on your machine and tell me what it says?  I'm surprised you don't get errors when building .o or executable files from generated .ll files.

pschachte: Wait, nevermind.  I just realised that that's one of the things we mask for final dump tests.

PR #525, opened 2025-12-08
Use proper LLVM structs for closures

jimbxb:
This always bothered me that I could have done better. Here is that better.

The has the benefit of:
- minor reduction in the allocated size of closures with closed params, due to struct packing
- producing more idiomatic LLVM


pschachte (approved):
Nice.  The generated LLVM looks a lot nicer.

PR #526, opened 2025-12-10
Add lazy(T) type to the stdlib

jimbxb:


pschachte:
I think your first approach was better.  I didn't notice you had separate slots for the thinker and the thought.  That should work fine, and be significantly more efficient, as it avoids another allocation.

PR #529, opened 2025-12-12
feat(#495): add foreign lpvm sizeof instr

### LINKED ISSUE #495: Add a language primitive to return the immediate size of data
Something like `sizeof` in C, except it should probably return the size in bits.  Like `sizeof`, it should be computed at compile-time.  For most algebraic types (declared with a `constructors` declaration), the size will be the size of a machine word, but for primitive types or types declared with only nilary constructors, it may be smaller.

It should probably take a type name or type variable as input, because one would often want to use it where they only have the type, and not a value of that type.  That may mean it will need a special syntax.  For example, currently the `array` type allocates a memory block of *n* words for an array of size *n*, regardless of the type of the elements, when it could do a lot better for some types if it knew the size of the elements.

jimbxb:
Closed #495

### DIFF
--- a/WYBE.md
+++ b/WYBE.md
@@ -2442,6 +2442,13 @@ converting its type from *type1* to *type2*, without changing the
 representation. This just allows getting around LLVM strong typing; it does not
 actually require any instructions.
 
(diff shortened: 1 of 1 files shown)


pschachte:
Is this obsolete now?  The PR I just merged included the sizeof primitive.

PR #532, opened 2025-12-27
fix(#324): expand generic types when inlining

### LINKED ISSUE #324: Type checking leaves fully determined types generic in generated code
The function
```
pub def even_primes:list(int) = [2]
```
generates the code
```
even_primes(?#result##0:wybe.list(wybe.int))<{}; {}>:
  AliasPairs: []
  InterestingCallProperties: []
    foreign lpvm alloc(16:wybe.int, ?tmp#4##0:wybe.list(T)) @list:9:27
    foreign lpvm mutate(~tmp#4##0:wybe.list(T), ?tmp#5##0:wybe.list(T), 0:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 2:T) @list:9:27
    foreign lpvm mutate(~tmp#5##0:wybe.list(T), ?#result##0:wybe.list(wybe.int), 8:wybe.int, 1:wybe.int, 16:wybe.int, 0:wybe.int, 0:wybe.list(T)) @list:9:27
```
Note that the proc type is correct, but after that, all the `list(int)` types show up as `list(T)`.

jimbxb:
Closes #324

### DIFF
--- a/test-cases/complex/exp/testcase_multi_specz-drone.exp
+++ b/test-cases/complex/exp/testcase_multi_specz-drone.exp
@@ -70,9 +70,9 @@ module top-level code > {terminal,inline,impure} (0 calls)
     foreign lpvm store(0:wybe.phantom, <<wybe.io.io>>:wybe.phantom) @io:nn:nn
     foreign c {impure} gc_init @memory_management:nn:nn
     drone.<0><{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}> #1
-    foreign lpvm alloc(16:wybe.int, ?tmp#9##0:wybe.c_array(T)) @command_line:nn:nn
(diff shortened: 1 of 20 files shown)


src/BodyBuilder.hs
```diff
@@ -217,11 +217,17 @@ type ComputedCalls = Map Prim [PrimArg]
 -- |Allocate the next temp variable name and ensure it's not allocated again
 freshVarName :: BodyBuilder PrimVarName
 freshVarName = do
-    tmp <- gets tmpCount
+    tmp <- freshTmp
     logBuild $ "Generating fresh variable " ++ mkTempName tmp
```
pschachte: I think it would be cleaner to call mkTempName in a let, so it isn't repeated.

PR #540, opened 2026-02-19
Add explicit size int and count types (8, 16, 32, and 64 bits)

### LINKED ISSUE #494: Remove the +=, *=, etc. library operations
and replace their uses with the new f(!variable), eg, replace !x += 5 with !x + 5.  Then look for places in the library where ?var = f(..., var, ...) can be replaced with f(..., !var, ...).

pschachte:
Add a test case for all the operations of these types.  Fix a few bugs discovered.
Also close #494, removing the +=, -=, etc. operations, since they're now redundant.


pschachte:
Interesting point about macros.  In general, I'd rather allow users to write normal Wybe code and have the compiler specialise it for different cases.  That's especially true in this case, given how similar the code for the different versions is (the only differences are the representation, the upper and lower bounds, and the specialised C functions called).  But I really don't see how to do that in this case.  You'd kind of want a type that's parametric in its bitwidth, and we only allow types to be parameterised by other types.

PR #541, opened 2026-02-23
Add test for endianness, and formatting with radix  to standard library

pschachte:


wybelibs/wybe/machine_word.wybe
```diff
@@ -5,8 +5,15 @@ pragma no_standard_library # Standard library can't depend on itself!
 
 representation is address
 
-use wybe.int
+use wybe.int, wybe.count
 
-pub def word_size_bytes:int = foreign lpvm sizeof(_:_)
+pub def {inline} word_size_bytes:int = foreign lpvm sizeof(_:_)
```
jimbxb: I'd have hoped this proc was already small enough to inline. It should just compile to a `move(8, ?result)`

PR #548, opened 2026-03-03
Add 4th benchmark program:  word count

pschachte:


.gitignore
```diff
@@ -60,3 +60,9 @@ stack.yaml.lock
 LOG*
 src/c_config
 src/CConfig.hs
+benchmarks/wc/wybe_wc
```
jimbxb: Don't these get cleaned automatically? If not, can they be?

pschachte: I'd rather not automatically clean up the executables.  It's useful to have them around to experiment with.

PR #550, opened 2026-03-19
feat(#545): box/unbox large primitives when cast to a type variable

### LINKED ISSUE #545: Large primitive types should be boxed/unboxed when passed as a generic value
See: https://github.com/pschachte/wybe/issues/543#issuecomment-3963903156_

POC: todo

jimbxb:
Closes #545


jimbxb:
Upon further consideration, I think I should move this transformation to the unbrancher.

This same sort of transformation for higher order calls, and inside the closure prices. That's because all arts to higher order calls have a similar zext/trunc that is a bug, too. These procs are only properly available after unbranching, so it's best we move all of the boxing/unboxing instead.

This does get a little tricky on the caller side. Ideally we generate the box/unbox in the unbrancher when calling a proc, but if that call is ever inlined, then the box may not be required. I'll play around and see if I can get this working.

---

As it stands, boxing will only automatically happen for relatively larger types. I don't think it makes sense to automatically generate these for smaller types, so I will not be generating them if not required.

I'd also like to open up these LPVM instructions to the user. I don't see any reason not to, though the size instruction could be a footgun we could just infer for the user. Type checking is relatively simple -- if present, the size argument must have an integral type, the boxed argument must have a variable type, and the unboxed can have any type.

pschachte:
I would hope that the programmer wouldn't need to think about boxing at all, except maybe across the foreign interface.

I think Wybe mostly handles >64 bit primitive types correctly right now.  LLVM handles them correctly in procedure calls (presumably by passing them on the stack, or perhaps across multiple registers).  I think storing them as arguments in an algebraic type should be fine, too, as long as LLVM can handle store and load instructions of more >64 bits.  AFAIK, it only goes wrong when a value of a large type is read from or written to a variable of generic type.  It always needs to happen then, for any type greater than word size, and is never needed otherwise.

I think I see what you mean about args in higher order calls.  Those are often generically typed.  But they should only be an issue when they _are_ generically typed.  If you call a HO proc of type, say, (int128, ?float128), is there any reason you can't just pass it a 128 bit int and have it return a 128 bit float?  I'm assuming LLVM will handle that correctly.

I don't think we need a new LPVM instruction to box or unbox big values.  It should be enough to just use an alloc followed by a mutate, and an access, respectively.  Unless there's something gained by making instructions for it.  This also has the advantage of automatically benefiting from Saumitra's work on escape analysis and transforming non-escaping alloc instructions into stack allocations.

Am I missing something in all this?

jimbxb:
I'd not considered just using alloc/etc to implement this. That would make it considerably simpler and amenable to other optimisations automatically.

---

Higher-order calls treat all arguments as untyped. This was to allow a higher order term be used in a glass a variably typed HO term was expected, eg

```
def do_something(f:(X)) { pass }

?f:(int) = { _ = @:int }
do_something(f)
```

Due to this, we need to box/unbox all higher order arguments as well.

pschachte:
I see.  OK, so the trampoline will need to handle the boxing/unboxing, right?  If so, then wouldn't the generated trampoline have generic argument types, and pass those to procs with specific types?  If so, then I would think that the generic <-> concrete type conversions, which would implement boxing/unboxing, would handle this when the trampoline is compiled.

I don't see why this should be done during unbranching.  That code is messy enough.  Since this is related to types, doing this during type checking (really it would be mode checking, because you don't know the direction of flow until then, and type checking doesn't actually change any code) seems like the best fit.

jimbxb:
Exactly. The trampoline should be responsible for the boxing/unboxing.

I need to check the implementation, but I don't think mode checking knows about trampolines, and they're created during unbranching

jimbxb:
Okay, that should be it refactored for the first order case. Major change is confined to Unbranch. There is a lot going on, but most of it is just trying to bubble up the pre/post stmts from when you compile the args.

I will tackle the higher order case in time, but this part should be ready now

pschachte (approved):
LGTM

pschachte:
Uh-oh.  There was one merge conflict, and I took a crack at resolving it, but I must have messed up, because it no longer compiles.  Can you please fix this, so I can merge it?

jimbxb:
No worries. I'll roll it back and rebase.

Before we merge, I'd like to fix the HO case as well. Better to wrap it up now than have it another `nyi`

src/Expansion.hs
```diff
@@ -448,10 +449,6 @@ inputOutputParams proto =
 addOutputNaming :: OptPos -> (PrimParam,PrimArg) -> Expander ()
 addOutputNaming _ (param@(PrimParam pname ty FlowOut _ _),
                      v@ArgVar{argVarName=vname}) = do
-    when (AnyType == ty) $
-      shouldnt $ "Danger: untyped param: " ++ show param
-    when (AnyType == argType v) $
-      shouldnt $ "Danger: untyped argument: " ++ show v
```
jimbxb, line 454: This is "okay" now, since closures now have untyped params

jimbxb:
Okay, so I'm a little disappointed, but I had to weaken a HO lowering optimisation.

When expanding, you may find that a you know which proc a HO call is calling. In this case previously, we could safely call the non-trampoline variant of that proc, as the trampoline proc had no extra instructions inside it (just a call to the non-trampoline).

Now, since we may have some box/unbox instructions on the caller and caller sides of the trampoline, we can't just replace the call with a call to the non-trampoline variant, and can only reduce the call to a first-order call to the trampoline. Further inlining would then correctly inline the trampoline variant and unbox/box as required.

We can strengthen this to not apply when there's no boxing involved, if required. I'm not exactly sure how we could recognise that, but we could just check if any needed args don't need to be boxed.

---

I would really like to make the env unmarshalling explicit inside the trampoline, also. For that, though, there is the complication that some of the free parameters may become unneeded, and so we can't just use an `lpvm access` instruction. The offset is also a little hard to calculate because we rely on LLVM's struct packing for closure structs.

I'm sure you'd not be too happy, but I'm thinking that a `lpvm unmarshall(env, ?free1, ?free2, ...)` could be "useful" here. It would have the semantics of a series of `getelementptr; read` when called inside a trampoline. When the trampoline is inlined in the HO lowering case, the call to `unmarshall` should have the `env` argument as a reference to an `argclosure` and can directly assign the `?free`s to the closed primarg.

If you have any better ideas I'd be happy to hear you out!

jimbxb:
Turns out recognising that was pretty simple if we bake it into the construction of the trampoline!

pschachte:
So, just to confirm, is this ready to be merged now?

jimbxb:
Yes

PR #551, opened 2026-03-22
Add Escape Analysis

SaumitraLohokare:
This PR implements intraprocedural escape analysis for LPVM alloc instructions. Allocations whose result does not outlive the current procedure call are now emitted as LLVM alloca (stack) instead of heap allocations, avoiding GC pressure for short-lived local structs.

### How it works
Two checks are combined at each alloc site:

1. Mutation-chain reachability - backwards fixed-point over the body. A variable escapes if it is an output parameter, or reaches one via mutate or pass-through edges.

2. Alias-map check - checks whether the alloc result is connected to a global or aliased output in the incremental alias disjoint-set.

If neither fires and the size is a compile-time constant, the `{stack}` flag is added to the alloc. LLVM codegen then lowers it to `alloca i8, i64 <size> + ptrtoint` instead of a GC malloc call.

### Verification

```py
pub type point { pub point(x: int, y: int) }

# mid is never returned → should be stack-allocated
pub def print_midpoint(p1: point, p2: point) use !io {
    ?mid = point((p1^x + p2^x) / 2, (p1^y + p2^y) / 2)
    !println(mid^x)
    !println(mid^y)
}

# q is an output parameter → heap-allocated
pub def make_point(?q: point) {
    ?q = point(10, 20)
}
```

**LLVM:**

`make_point`: Heap-allocated.
```ll
define external fastcc i64 @"escape_ex#.make_point<0>"() {
  %"tmp#6##0" = call ccc ptr @wybe_malloc(i32 16)   ; heap allocation
  %"tmp#3##0" = ptrtoint ptr %"tmp#6##0" to i64
  %"tmp#7##0" = inttoptr i64 %"tmp#3##0" to ptr
  store i64 10, ptr %"tmp#7##0"                      ; x = 10
  %"tmp#8##0" = add i64 %"tmp#3##0", 8
  %"tmp#9##0" = inttoptr i64 %"tmp#8##0" to ptr
  store i64 20, ptr %"tmp#9##0"                      ; y = 20
  ret i64 %"tmp#3##0"
}
```

`print_midpoint`: Stack-allocated.

```ll
define external fastcc void @"escape_ex#.print_midpoint<0>"(i64 %"p1##0", i64 %"p2##0") {
  ; compute (p1.x + p2.x) / 2
  ...
  ; compute (p1.y + p2.y) / 2
  ...
  ; mid = point(...) — stack allocated
  %"tmp#44##0" = alloca i8, i64 16, align 8          ; ← alloca, not malloc
  %"tmp#25##0" = ptrtoint ptr %"tmp#44##0" to i64
  %"tmp#45##0" = inttoptr i64 %"tmp#25##0" to ptr
  store i64 %"tmp#1##0", ptr %"tmp#45##0"            ; mid.x
  %"tmp#46##0" = add i64 %"tmp#25##0", 8
  %"tmp#47##0" = inttoptr i64 %"tmp#46##0" to ptr
  store i64 %"tmp#5##0", ptr %"tmp#47##0"            ; mid.y
  ; read back and print
  ...
  ret void
}
```


src/LLVM.hs
```diff
@@ -941,11 +941,28 @@ writeLLVMCall op flags args pos = do
...
+        ([sz],[out]) ->
+            if "stack" `elem` flags
+            then case argIntVal sz of
+                -- Stack-allocate: emit alloca and cast the pointer to i64.
+                -- Requires a constant size (guaranteed by escape analysis).
+                -- Side-effect: any call that receives this address as an i64
+                -- argument cannot be a tail call (the callee would dereference
+                -- a pointer into a stack frame that tail-call would have freed).
+                -- LLVM enforces this automatically; sets doesAlloca accordingly.
+                Just sizeVal -> do
```
pschachte, line 1017: Does LLVM's alloca only work for fixed size allocations?  If so, please say that here in a comment.  If not, please add a comment saying this is a temporary restriction.  Unless you can generalise it so it works for variable sized allocations.

src/Transform.hs
```diff
@@ -70,6 +70,84 @@ initAliasMap proto speczVersion = do
...
+            ++ concatMap collectAllBodyPrims (Maybe.maybeToList dflt)
+
+
+-- | Returns True if this prim is lpvm alloc or lpvm mutate.
+isAllocOrMutate :: Prim -> Bool
+isAllocOrMutate (PrimForeign "lpvm" "alloc"  _ _) = True
+isAllocOrMutate (PrimForeign "lpvm" "mutate" _ _) = True
+isAllocOrMutate _                                  = False
+
+-- | Compute the set of variables whose allocated memory must survive the
```
pschachte: I think you mean "may" rather than "must".  This is an important distinction:  this optimisation can only be done if it's _impossible_ for the allocation to survive the proc body it appears in.  You have to be conservative in approximating this (and you have to approximate; you can't always get it exactly right).

pschachte: Oh, I think I understand what you mean.  OK, it's difficult to express this clearly, but I think it's clearer to write of what will happen when the program runs, so think of a value as escaping a proc body if it is possible that the value could be referred to after the proc returns.

src/Transform.hs
```diff
@@ -70,6 +70,84 @@ initAliasMap proto speczVersion = do
...
+computeEscapedVars :: PrimProto -> ProcBody -> Set PrimVarName
+computeEscapedVars proto body =
+    let escaped0 = Set.fromList
+                    [ primParamName p
+                    | p <- primProtoParams proto
+                    , isOutputFlow (primParamFlow p) ]
+        prims = List.map content $ collectAllBodyPrims body
+        -- For mutate(fIn, fOut, offset, destr, size, startOff, member, ...),
+        -- if fOut escapes then:
+        --   (a) fIn escapes (same struct, just a new version)
```
pschachte: fIn might not escape in this case, if the destructive flag is not set in the mutate.  In that case, the mutate instruction makes a fresh copy, so fOut escaping doesn't imply that fIn does, unless the structure contains pointers.

If your analysis/transformation is happening before the destructive update transformation, it would be useful to document that this limits the stack allocation transformation here.  Otherwise, I think you should be able to do better here.

src/Transform.hs
```diff
@@ -70,6 +70,84 @@ initAliasMap proto speczVersion = do
...
+        --   (b) any field value (member) stored into it also escapes,
+        --       because that pointer will be embedded in the escaping struct.
+        mutateEdges = [ (argVarName fOut, argVarName vin)
+                | PrimForeign "lpvm" "mutate" _ args@(fIn:fOut:_) <- prims
+                , argIsVar fIn
+                , argIsVar fOut
+                , vin <- fIn : [ m | m <- List.drop 6 args, argIsVar m ] ]
+        -- For any non-alloc/non-mutate instruction, propagate escape backward:
+        --   • foreign ops (e.g. llvm or for constructor tagging, llvm move,
+        --     lpvm cast): type-match only — an input escapes iff an output of
```
pschachte: This only true for foreign llvm and lpvm (other than mutate) calls.  A foreign C call could store a value in a C structure, which could be accessed by another C function.  You should treat C (or, for generality, any language other than llvm and lpvm) calls at least as conservatively as Wybe calls.

pschachte (changes requested):
This looks great, and I really appreciate your documentation here.  Makes the code much easier to review.

I've placed a few requests for documentation clarification inline, and one suggested improvement.

The other thing for you to fix is all the failing test cases.  Before resubmitting, please run the update-exp script and scroll carefully through the differences in each file, making sure that they are all as expected (they're mostly calls to foreign lpvm alloc that have had a `{stack}` flag added, and the consequent changes to LLVM code).  I.e., make sure that the all those stack allocations are safe.  For each test case, if so, say yes, and it will replace the .exp file with the new output.  Then add all the changed .exp files to the pull request branch.  If not, you'll need to fix the bugs.

You should also add your new test case to the test-cases/final-dump directory and run update-exp; it will show you the output and prompt you to accept or reject it.  Then also add the source and .exp files to the branch.  Finally, please add a couple more tests to that test case, one that creates a fresh structure, passes it to another proc, which passes it back, and see if the escape analysis still sees that it doesn't escape (it's OK if it doesn't, but we should know), and one that does something similar, but the structure really does escape, so we can see that it doesn't optimise cases it shouldn't.

src/Transform.hs
```diff
@@ -70,6 +70,84 @@ initAliasMap proto speczVersion = do
             ) emptyDS inputParams
 
 
+-- | Collect all prims from a body, including all fork branches (conservative).
+collectAllBodyPrims :: ProcBody -> [Placed Prim]
```
jimbxb: Can this be done using one of the fold methods implemented in AST?

src/Transform.hs
```diff
@@ -176,11 +258,32 @@ transformPrim callSiteMap (aliasMap, deadCells) prim = do
...
+                        ++ " | escapedByMutation=" ++ show escapedByMutation
+                        ++ " | constSize=" ++ show constSize
+                        ++ (if not escaped && constSize
+                            then " => stack-allocate"
+                            else " => heap-allocate")
                 let primc' = case result of
-                        Nothing -> primc
+                        Nothing ->
+                            if not escaped && constSize
+                            then PrimForeign "lpvm" "alloc" ("stack":flags) args
```
jimbxb: We should skip this if the `stack` flag is already present

src/LLVM.hs
SaumitraLohokare: `alloca` also works for variable sized allocations from my research, but I'm not super familiar with LLVM in general so I chose to take a safer approach. I'll comment this 👍 Thank you

SaumitraLohokare:
I also had the thought of adding a flag to disable/enable escape analysis. Is this a good idea or should we just leave it as always enabled?

pschachte:
Adding a command line flag is a good idea.  It is stack allocation, not the escape analysis, that would be relevant to programmers, so I would name the switch in terms of stack allocation rather than escape analysis.  So far, we optimise by default, so I think the switch should disable automatic stack allocation.

src/Transform.hs
SaumitraLohokare: I used `foldBodyPrims` for this. Please let me know if this is fine.

SaumitraLohokare:
I added the flag for `stack-alloc` under `--opt`. It is on by default. And also added a new test case for stack allocation.
Ran `update-exp` as well, all tests passing.

src/LLVM.hs
```diff
@@ -941,11 +941,28 @@ writeLLVMCall op flags args pos = do
...
-        ([sz],[out]) -> heapAlloc out sz pos
+        ([sz],[out]) ->
+            if "stack" `elem` flags
+            then case argIntVal sz of
+                -- Stack-allocate: emit alloca and cast the pointer to i64.
+                -- Requires a constant size (guaranteed by escape analysis).
+                -- Side-effect: any call that receives this address as an i64
+                -- argument cannot be a tail call (the callee would dereference
+                -- a pointer into a stack frame that tail-call would have freed).
+                -- LLVM enforces this automatically; sets doesAlloca accordingly.
```
jimbxb: We should be able to do a tail call if the stack allocated space never enters the tail call.

src/LLVM.hs
jimbxb: I don't see why we can't support variably sized stack allocations :)

src/Transform.hs
```diff
@@ -70,6 +71,86 @@ initAliasMap proto speczVersion = do
...
+computeEscapedVars proto body =
+    let escaped0 = Set.fromList
+                    [ primParamName p
+                    | p <- primProtoParams proto
+                    , isOutputFlow (primParamFlow p) ]
+        prims = collectAllBodyPrims body
+        -- For mutate(fIn, fOut, offset, destr, size, startOff, member, ...),
+        -- if fOut escapes then:
+        --   (a) fIn escapes (same struct, just a new version).  Note: this
+        --       analysis runs before the destructive-update transformation sets
```
jimbxb, line 262: Why can't we run it afterwards?

src/Transform.hs
```diff
@@ -70,6 +71,86 @@ initAliasMap proto speczVersion = do
...
+                , argIsVar fIn
+                , argIsVar fOut
+                , vin <- fIn : [ m | m <- List.drop 6 args, argIsVar m ] ]
+        -- For any non-alloc/non-mutate instruction, propagate escape backward:
+        --   • foreign llvm/lpvm ops (e.g. constructor tagging, llvm move,
+        --     lpvm cast): type-match only — an input escapes iff an output of
+        --     the same type escapes.  This is sound because these ops can't
+        --     embed a value into a result of a different type.
+        --   • user-defined calls (PrimCall) and foreign calls to other
+        --     languages (e.g. C): we have no visibility into the callee, so
```
jimbxb: I would hope for Wybe procs we do.

Is there a way we can annotate Params as saying "this input escapes iff some output also escapes"? 

Not sure that makes sense, but I'm sure we can do better here. For (co-recursive) price we'd have to be conservative or attempt to find a fixed point.

jimbxb (changes requested):

src/LLVM.hs
jimbxb: Something like

```
def foo use !io {
    ?p = position(1, 2)
    !print(p)
    !foo
}
```
Should be able to stack allocate `p`. 

I'm thinking this case could be fairly common, say if something is used for an intermediate computation before a tail call.

src/Transform.hs
jimbxb: For instance,

```
def foo(p0, ?p1, q0, ?q1) {
    p0 = ?p1
    q0 = ?q1
}
```
For a call to `foo`, `p0` should only escape if `p1` does, not if `q1` does. 

Unless I'm misunderstanding, I believe the current analysis would have `p0` and `q0` escape if `p1` or `q1` escape.

src/LLVM.hs
SaumitraLohokare: Hello, sorry for the late reply. Been a bit busy with proposal and presentations. But I looked into variable sized alloca and I think it isn't worth the effort. Mainly for two reasons:

- Variable-sized alloca forces LLVM to keep a frame pointer because the stack frame size can't be computed statically. Wybe actively uses TCO (TailCallModCons is an optimization flag). These two features are fundamentally incompatible.
- `malloc` can return null; a stack overflow from a large VLA just segfaults or corrupts the stack silently. For objects that are large or runtime-sized, heap allocation is the safer default.

src/Transform.hs
SaumitraLohokare: You are correct. I did try to implement a fix for this, but my fix is causing `./complex/exp/testcase_multi_specz-int_list.out` to fail with a segmentation fault.
Since this is more of a precision issue, rather than correctness, I'd like to move this to my next PR so I can take a closer look at this.

src/Transform.hs
SaumitraLohokare: technically we could, but it would require splitting the current single-pass design. The destr flags are set incrementally during `transformBody` using an alias map that's built as the body is traversed; `computeEscapedVars` runs before that traversal starts, so the updated flags don't exist yet. To use post-destr escape analysis we'd need a preliminary sub-pass just to accumulate the alias map and annotate destr flags, then re-run escape analysis on the result. Happy to do that as a follow-up if the precision loss is worth fixing.

src/LLVM.hs
SaumitraLohokare: Good point! Fixed - instead of blocking all tail calls once any `alloca` is emitted, we now track which variables hold stack-allocated addresses and only block tail calls that directly receive one of those variables as an argument. Since escape analysis already guarantees stack-allocated addresses can't reach a call through any other mechanism, checking direct arguments is sufficient.
This adds back a lot of tail calls.

src/LLVM.hs
jimbxb: I think we should support it in LLVM generation, but never generate it in the escape analysis. There are cases where we alloc a variable length array, and if the user wanted to supply their own `stack` flag I don't see why we shouldn't support that

test-cases/final-dump/stack_alloc.wybe
```diff
@@ -0,0 +1,38 @@
...
+
+# Test 2: pass-through — the position is passed to another proc which returns
+# it unchanged, and the result is used locally (never returned by this proc).
+# Expected: the alloc should still be stack-allocated, since neither p nor q
+# is an output parameter of test_pass_through.  The escape analysis should
+# see that the conservative edge from pass_back does not fire (q is not
+# escaped), so p also does not escape.
+def {noinline} test_pass_through use !io {
+    ?p = position(1, 2)
+    !pass_back(p, ?q)
```
jimbxb: Why does this have a `!`? It shouldn't need it

src/LLVM.hs
pschachte: Since tail recursion optimisation only works for certain calling conventions, and from what you've told me, you can't use a calling convention that supports TRO together with variable-sized alloca, I think we should call this Not Yet Implemented for now (there's a `nyi` function to report such missing features).  If we get around to doing our own TRO during LLVM generation, we should revisit this.

src/Transform.hs
jimbxb: Sure. Ill resolve this for now and hopefully won't forget about it

src/Transform.hs
jimbxb: Again, I'll hopefully not forget this.

test-cases/final-dump/stack_alloc.wybe
SaumitraLohokare: Fixed 👍

jimbxb: My concern is that this passed the type checker. I'd have thought it would cause a build failure on this test case. 

Can you please raise an issue if this doesn't cause an error?

jimbxb: Oh, I see. The expected output _did_ have the warnings. All good

src/LLVM.hs
jimbxb: @SaumitraLohokare please replace with `nyi`

test-cases/final-dump/stack_alloc.wybe
```diff
@@ -0,0 +1,38 @@
+use position
```
pschachte, line 1: This test code doesn't actually test what it's meant to test, because the compiler's constant factoring turns the structures you use into static constants.  Can you fix this to pass in, say, the y component of the positions to build, to prevent the compiler from statically allocating them?

pschachte (changes requested):

src/LLVM.hs
```diff
@@ -2184,7 +2221,11 @@ stackAlloc :: PrimArg -> Int -> LLVM ()
 stackAlloc result size = do
     llvmAssignResult result $ "alloca i8, i64 " ++ show size
         ++ ", align " ++ show wordSizeBytes
-    modify $ \s -> s { doesAlloca = True }
+    case result of
```
jimbxb: Can you please refactor this and the other instance where we add to this map into a separate function?

src/LLVM.hs
```diff
@@ -876,18 +876,27 @@ writeHOCall closure args pos = do
...
+-- being passed to the call; if any directly reference a stack-allocated
+-- variable, we cannot emit a tail marker (the callee would dereference a
+-- pointer into a stack frame that tail-call optimisation would have freed).
+-- If no input arg references a stack-allocated variable, TCO is safe even when
+-- the current body contains alloca instructions, because the escape analysis
+-- guarantees those allocations do not reach the callee.
+tailMarker :: Bool -> [PrimArg] -> LLVM String
+tailMarker must ins = do
+    stackVars <- gets stackAllocedVars
+    let passesStackVar = any (\arg -> case arg of
```
jimbxb, line 926: I'm not 100% certain this is correct. The map contains vars that are stack allocated, but I believe another variable _could_ also have that pointer despite not being explicitly stack allocated through a move?

jimbxb: Something like this

```
def loop(p0:position, depth:int) use !io {
    ?p1 = position(1, p0^y)
    !printPosition(p0)
    pass_back(p1, ?p2)
    if { depth > 0 :: 
        !loop(p2, depth - 1)
    }
}

def {noinline} pass_back(p:position) = p
```

`p1` should be stack allocated (it doesn't escape from `loop`, and `pass_back` should conservatively infer that `p2` doesn't escape either), yet there is no explicit `{stack} alloc` of `p2` and hence I don't think this should be an LLVM `tail` call.

src/Transform.hs
pschachte: I think this is not so much an issue of escaping as one of aliasing.  It becomes more interesting from the standpoint of a proc that calls foo.  Eg,

```
def bar(?z) {
    ?x = position(3,5)
    foo(x, ?y, x, ?z)
}
```

Here it's important to know that x *does* escape bar.

src/LLVM.hs
pschachte: If there's a choice between TRO and stack allocation, I think we should pick TRO, since TRO allows a proc to run in fixed stack space.  Saumitra has updated his implementation to disable stack allocation for procs where TRO should apply.

jimbxb: I agree, we should prefer TRO to stack allocation in general.

I don't follow though. I think this code is doing the opposite -- disabling TRO when a stack allocation occurs.

src/LLVM.hs
jimbxb, line 964: Sorry if I'm just blind to this, but shouldn't we also propagate the stack-alloc-ness here for all outputs? At the moment, the propagation appears to only happen for LPVM calls, but what happens if the stack allocated variable gets passed to a wybe proc?

jimbxb: Hmm I see. I think we are being super conservative now and we always assume a value passed in escapes. Unfortunate

test-cases/execution/stack_alloc_escape.wybe
```diff
@@ -0,0 +1,22 @@
...
+# allocation here, combined with tail-call optimisation of the continuations,
+# would free the cells while they are still in use (a use-after-free).
+#
+# This is an *execution* test on purpose: the miscompilation only shows up at
+# runtime (garbage output), not in the dumped IR.
+
+def show(n:int) use !io {
+    ?xs = []
+    ?i = n
+    # If while is a new procedure, then why is it still trying to stack allocate?
```
jimbxb: I don't understand this comment

test-cases/execution/stack_alloc_escape.wybe
```diff
@@ -0,0 +1,22 @@
+# Regression test for stack-allocation escape analysis.
+#
+# Build a list locally in a loop, then consume it locally.  After unbranching,
+# the cons cells are allocated across tail-recursive continuation frames and
+# only read by a later continuation (the for-loop), so they outlive the frame
+# that built them.  They must therefore NOT be stack-allocated: stack
+# allocation here, combined with tail-call optimisation of the continuations,
+# would free the cells while they are still in use (a use-after-free).
+#
+# This is an *execution* test on purpose: the miscompilation only shows up at
```
jimbxb, line 10: I think it would be useful to have this as a inal-dump test also, to ensure we are actually stack allocating?

test-cases/execution/stack_alloc_escape.wybe
SaumitraLohokare: Sorry, I was just typing my thoughts out, missed this one.

test-cases/execution/stack_alloc_escape.wybe
SaumitraLohokare: Yup, will do.

src/LLVM.hs
SaumitraLohokare: I see why it reads that way, but the current code doesn't disable TRO on stack allocation - that was the old `doesAlloca` boolean (any alloca killed every tail call in the body). `tailMarker` is now per-call and only drops the marker if a stack address flows into that call's args. And that never happens for a real call: `callArgEsc` already marks any pointer passed to a `PrimCall`/`PrimHigher`/`foreign call` as escaping, so it's heap-allocated, not stack. So a value that could be TRO'd is never a stack candidate to begin with - same outcome @pschachte described, just done from the escape-analysis side. TRO wins.

src/Transform.hs
```diff
@@ -139,69 +272,148 @@ transformForks caller body (aliasMap, deadCells) callSiteMap = do
...
+                -- outVar is created *by* this alloc, so it cannot yet be
+                -- connected to any global/param in that map.  Empirically it is
+                -- False for every alloc across the whole test suite.  The
+                -- authoritative, sound check is escapedByMutation
+                -- (computeEscapedVars), which scans the whole body and so sees
+                -- escapes that occur after the alloc (the usual case: build a
+                -- value, then store/return it).  escapedByAlias is kept only as
+                -- a cheap, order-dependent early check that can add escapes but
+                -- never remove them, so it cannot affect soundness.  See §4/§9
+                -- of escape_analysis.md.
```
jimbxb: What's this file?

test-cases/execution/stack_alloc_global_escape.wybe
```diff
@@ -0,0 +1,39 @@
...
+# Overwrite the previously-used stack region with unrelated data.
+def {noinline} clobber(n:int, ?out:int) {
+    ?acc = 0
+    for ?i in 1..n { ?acc = acc + i }
+    ?out = acc
+}
+
+!stash(42)
+clobber(1000, ?junk)
+use saved in {
```
jimbxb, line 35: Any reason for the use block?

test-cases/execution/stack_alloc_reuse_escape.wybe
```diff
@@ -0,0 +1,43 @@
...
+# escape analysis) makes the two field writes update `c` in place, so the
+# final version of `c` *is* the stack memory.  After `c^a` is read, `c` is
+# dead, and dead-cell reuse offers its memory to the next alloc — the cell
+# `d`, which is then stored into the global `saved` and therefore ESCAPES.
+#
+# Reusing `c`'s stack memory for the escaping `d` would leave the global
+# pointing into a torn-down stack frame (use-after-free).  The fix forbids
+# reusing a stack-allocated dead cell: `d` must be heap-allocated instead.
+#
+# This is an *execution* test on purpose: with the bug the dumped IR merely
```
jimbxb, line 15: Again, I think this is a good candidate for a final dump and execution test. There's comments in this file that state something should be stack allocated, but the final dump .exp doesn't really help us verify that :(

Maybe we can have a symlink for this in the final-dump directory, that way we don't have to keep them in sync? I assume git would handle that all pretty well.

src/LLVM.hs
pschachte: @SaumitraLohokare to clarify James's point, please use the function `nyi` (not yet implemented) in place of `shouldnt` to report things that are errors due to limitations of the current implementation, rather than inherent errors.  That would apply to the `shouldnt` below about non-constant size alloc.

src/Transform.hs
pschachte: If you've written up a separate document, it would be best to embed it as a lengthy comment in a source file (probably this one), between `--  BEGIN MAJOR DOC` and `--  END MAJOR DOC` comment lines.  Then do `make doc` in the top level directory.  That will update the `README.md` file in the `src` directory, which provides high-level documentation about the implementation.

src/Transform.hs
```diff
@@ -139,69 +272,148 @@ transformForks caller body (aliasMap, deadCells) callSiteMap = do
...
+                let escapedByAlias = isArgEscaped aliasMap outVar
+                let escapedByMutation = case outVar of
+                        ArgVar{argVarName=n} -> Set.member n escapedVars
+                        _                   -> True
+                let escaped   = escapedByAlias || escapedByMutation
+                -- NOTE: We only stack-allocate constant-sized allocs. LLVM does
+                -- support variable-sized alloca (for C99 VLAs), but it forces a
+                -- frame pointer, blocking tail-call optimisation. It also makes
+                -- --stack-alloc-limit unenforceable at compile time. In practice
+                -- this is not a limitation: Wybe types are always statically sized.
```
pschachte: That's not actually true.  Look at the `array` and `c_string` types.  It is true of types defined through constructors.

pschachte (changes requested):
A few small requested changes.

src/Transform.hs
SaumitraLohokare: That's correct, I've updated the comments to note this. Though the implemented logic is still correct (and unchanged).

SaumitraLohokare: I also made the `nyi` changes in the previous commit.

SaumitraLohokare: Regarding the `escape_analysis.md` doc, I'll push it by tonight (I'd like to have AI fix my grammar a bit 😅)

test-cases/execution/stack_alloc_reuse_escape.wybe
SaumitraLohokare: I've symlinked `stack_alloc_reuse_escape.wybe` into `final-dump/` and added the `.exp`.
the final-dump `.exp` now shows `c` as `{stack}` alloc and the escaping `d` as a plain heap alloc, so it actually verifies the stack-allocation claim in the comment, while the execution test still checks the runtime output.

pschachte (approved):
This looks fine now.  Let me know when you're ready for it to be merged.

pschachte:
... after you've fixed the merge conflicts.

src/Transform.hs
```diff
@@ -26,11 +26,125 @@ import           Data.Maybe    as Maybe
...
+-- Escape analysis is part of the **Transform** pass
+-- ([`transformProcBody`](/src/Transform.hs)), which runs after alias analysis has
+-- reached its fixed point. Transform already walks every procedure body to set
+-- the *destructive* flag on `mutate` instructions (the in-place-update
+-- optimisation). Stack allocation piggy-backs on that same walk.
+--
+-- For each (non-inline) procedure, Transform:
+--
+-- 1. Computes `escapedVars`, the set of variables that may escape — see
+--    [`computeEscapedVars`](/src/Transform.hs) (§5). This is computed **once**,
```
jimbxb, line 134: What do these `§`s reference?

SaumitraLohokare: The sections in this doc comment.

jimbxb: Right... there are two major docs.

I'm a bit overwhelmed by this all -- it seems awfully verbose and having snippets of compiler code throughout are just bound to go stale :/

pschachte: The verbosity is actually on me.  I discussed this with Saumitra, and felt that it's better to have this info than have to try to reconstruct later, so I suggested adding it as major doc.  You can always ignore it.  Yes, bit rot is always a danger.  Still, on balance, I think it's better to have it than not.

SaumitraLohokare:
Hello, I have fixed the build error. I believe it should build and pass all tests now.

test-cases/execution/stack_alloc_global_escape.wybe
jimbxb: @SaumitraLohokare please see this comment

SaumitraLohokare: No good reason. I assumed it was needed to bring the resource into scope, but module-level resources are already in scope at the top level. Happy to drop it.

pschachte (approved):

PR #552, opened 2026-04-01
Update doc to link to Software paper

pschachte:


jimbxb (approved):
Very nice to have that paper published. I'll have to read it now!

PR #553, opened 2026-04-16
feat: add unboxed sizeof

### LINKED ISSUE #542: sizeof gives wrong information for address representation types
The `address` representation type denotes a chunk of memory of unknown size at that address.  Therefore, the size of that type cannot be determined.  Currently, `foreign lpvm sizeof` returns 8 for that; it should instead report an error.

jimbxb:
Closes #542

### DIFF
--- a/src/Util.hs
+++ b/src/Util.hs
@@ -8,6 +8,7 @@
 module Util (sameLength, maybeNth, insertAt,
              setMapInsert, showArguments,
              fillLines, nop, sccElts, mapFst4, thd4, 
+             nextPowerOf2, prevPowerOf2, ceilDiv,
              DisjointSet, emptyDS, addOneToDS, unionTwoInDS,
              combineTwoDS, removeSingletonFromDS,
              addConnectedGroupToDS, removeOneFromDS,
@@ -123,9 +124,24 @@ mapFst4 :: (a0 -> a1) -> (a0, b, c, d) -> (a1, b, c, d)
 mapFst4 f (a, b, c, d) = (f a, b, c, d)
 
 
-
+-- Get the third of a 4-tuple
 thd4 :: (a, b, c, d) -> c
 thd4 (_, _, c, _) = c
+
+
+-- Find the next power of 2 no greater than the given integer
+nextPowerOf2 :: (Integral b, Integral a) => a -> b
+nextPowerOf2 n = ceiling $ logBase 2 $ fromIntegral n
+
+
+-- Find the previous power of 2 no lesser than the given integer
+prevPowerOf2 :: (Integral b, Integral a) => a -> b
+prevPowerOf2 n = floor $ logBase 2 $ fromIntegral n
+
+
+-- Ceiling'ed integer division
+ceilDiv :: Integral a => a -> a -> a
+ceilDiv a b = (a + b - 1) `div` b
 
 
 ----------------------------------------------------------------
@@ -394,4 +410,3 @@ createLocalCacheFile file = do
     srcFileHash <- _getFileHash file
     writeFile meta srcFileHash
     return cacheFile
-
--- a/test-cases/final-dump/sizeof.wybe
+++ b/test-cases/final-dump/sizeof.wybe
@@ -38,3 +38,9 @@ foreign lpvm {bits} sizeof(_:wider, ?wider_bits:int)
 foreign lpvm {bits, unboxed} sizeof(_:wider, ?wider_bits:int)
 !println("Wider bits (unboxed): $wider_bits")
 
+type wider_int is 128 bit unsigned {}
+
+foreign lpvm {bits} sizeof(_:wider_int, ?wider_int_bits:int)
+!println("Wider int bits: $wider_int_bits")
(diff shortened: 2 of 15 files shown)


src/Util.hs
```diff
@@ -123,11 +123,20 @@ mapFst4 :: (a0 -> a1) -> (a0, b, c, d) -> (a1, b, c, d)
...
+-- Get the third of a 4-tuple
 thd4 :: (a, b, c, d) -> c
 thd4 (_, _, c, _) = c
 
 
+-- Find the next power of 2 no greater than the given integer
+nextPowerOf2 :: (Integral b, Integral a) => a -> b
+nextPowerOf2 n = ceiling $ logBase 2 $ fromIntegral n
+
+-- Find the next power of 2 no lesser than the given integer
```
jimbxb: Oops. Typo

wybelibs/wybe/lazy.wybe
```diff
@@ -15,8 +15,10 @@ pub def {pure} get(lazy:_(T)):T = thought where {
     lazy = thunk(?thinker, ?thought)
     if { low_bit_unset(thinker) ::
         thinker(?thought)
-        foreign lpvm {impure} mutate(lazy, ?lazy, 0, 1, 16, 0, set_low_bit(thinker))
-        foreign lpvm {impure} mutate(lazy, ?lazy, 8, 1, 16, 0, thought)
+        ?thinker_size = foreign lpvm sizeof(thinker):int
+        ?total_size = thinker_size + foreign lpvm sizeof(thought):int
```
jimbxb: Whoops. I can use `{unboxed} sizeof` here!

test-cases/final-dump/sizeof.wybe
```diff
@@ -10,15 +12,29 @@ foreign lpvm {bits} sizeof(c, ?char_bits:count)
...
+
+type wider {
+    constructor wider(_, wide) | xxx | foo(c:char)
+}
+
+foreign lpvm {bits} sizeof(_:wider, ?wider_bits:int)
+!println("Wider bits: $wider_bits")
+foreign lpvm {bits, unboxed} sizeof(_:wider, ?wider_bits:int)
+!println("Wider bits (unboxed): $wider_bits")
+
```
pschachte, line 40: Could you add a test case finding the size of a 128-bit integer, both boxed and unboxed?

wybelibs/wybe/array.wybe
```diff
@@ -26,21 +26,22 @@ use wybe.bool, wybe.int, wybe.machine_word, wybe.list, wybe.c_array
 representation (T) is address 
 
 ## Accessor procedures backward compatible with original array type
-pub def length(a:_(T)):int = foreign lpvm access(a, 0, word_size_bytes, 0)
+pub def length(a:_(T)):int = 
+    foreign lpvm access(a, 0, foreign lpvm sizeof(_:T):int, 0)
```
pschachte: Shouldn't the length always be an int-sized value?  I don't see why it would be related to the size of the elements.

wybelibs/wybe/array.wybe
```diff
@@ -26,21 +26,22 @@ use wybe.bool, wybe.int, wybe.machine_word, wybe.list, wybe.c_array
...
 pub def raw_data(a:_(T)):raw_array =
-    foreign llvm add(a, word_size_bytes)
+    foreign llvm add(a, foreign lpvm sizeof(_:T):int)
```
pschachte, line 33: Same here, and below

wybelibs/wybe/array.wybe
```diff
@@ -149,7 +150,7 @@ pub type array_iterator(T) {
 # WARNING: does not initialise the array.
 # Allocate space for an array and record the size.
 def {inline} allocate_array(len:int,?ar:_(T),?size:int) {
-    ?size = (len + 1) * word_size_bytes
+    ?size = (len + 1) * foreign lpvm sizeof(_:T)
```
pschachte: This should be sizeof(int) + len * sizeof(T)

pschachte (changes requested):

wybelibs/wybe/array.wybe
jimbxb: Yes. Seems my dirty find-replace wasn't so sound

wybelibs/wybe/array.wybe
```diff
@@ -53,10 +54,10 @@ pub def array(len:int):_(T) = ar where {
 # Construct a Wybe array from a c_array.  This will need to copy the data.
 pub def array(data:c_array(T)):_(T) = ar where {
     allocate_array(data^length, ?ar, ?size)
-    ?i = 0
+    ?i = foreign lpvm sizeof(_:int):int
```
jimbxb, line 57: This zero was a bug. Good thing you prompted me to check @pschachte

test-cases/final-dump/sizeof.wybe
jimbxb: Prints 64 and 128 bits, respectively. Not sure thats the behaviour we want, but I think its okay

pschachte (approved):

PR #556, opened 2026-04-25
test: build sample and benchmarks in test suite

### LINKED ISSUE #539: Update the make test routine to at least build the sample and benchmark programs
...so that we detect bitrot.

jimbxb:
Closes #539, #557

Also:
- normalises some final-dump test source code locations
- streamlines some gitignores
- removes an old redundant final-dump .out (see commit e3790ca)


update-exp
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
jimbxb: Needed for fall through

update-exp
```diff
@@ -16,17 +16,21 @@ prefix () {
...
 
 
+force_accept=0
 # Determine whether to accept compiler output as expected.  Inputs are:
 #    source file name
 #    compiler output file name
 #    current expected output file name, or empty for new test cases
 agree () {
+    if [ $force_accept -ne 0 ] ; then
+        return 0
```
pschachte, line 40: I'm not sure 'force accept' is a good idea.  Sometimes (often!) going through all the changes is really tedious, and sometimes I go through accepting output at a glance, but I do at least glance at the changes.  It looks like 'force accept' means accept all following changes sight unseen, and that seems a bit hasty.

jimbxb: This option should always be used with caution, of course. It's only really useful if you're making small, sweeping changes that you're confident are correct, and then I do the sniff test in another program.

pschachte (changes requested):
OK, fair enough.  But could you put in a confirmation of the force command?  I don't generally like confirmations, but for something this drastic, I think a little caution is warranted.

update-exp
jimbxb: Should be fixed now.

update-exp
jimbxb: No longer required anyway.

pschachte (approved):

PR #563, opened 2026-05-26
 Closes #558 and #559

### LINKED ISSUE #558: Optimisation: Trim memcpy range when immediately overwriting the start or end of the destination
No point copying something that will only be overwritten!  For example, when using lpvm mutate on a cons cell to change the head (or tail, for that matter), we currently copy the whole cons cell, and then immediately overwrite the first (resp. last) word.  We actually only need to copy the word(s) that aren't about to be overwritten.

Where there are multiple mutate instructions in a row, we could trim the memcpy by all the initial and/or final memory that will be overwritten.

This will need to be done during LLVM generation.

### LINKED ISSUE #559: Optimisation:  replace small memcpy operations with individual loads and stores if that is faster
If the LLVM codegen already does this, don't bother!

This should be done after #558, so it applies more often

pschachte:
Closes #558
Closes #559

For non-destructive lpvm mutate, don't copy words that will be overwritten anyway.  Recognises cases where the data to be non-destructively written is a multiple of the machine word size, is word aligned, is at a compile-time constant offset from the start of the structure, and either:

  - it is at the beginning of the structure; or
  - it is at the end of the structure; or
  - the size of the structure is small (currently at most 5 machine words)

In the latter case, the LLVM compiler will generate inline load and store instructions to copy the data; in the first two cases, if the structure is larger, the compiler may generate a single loop to copy the structure.

Note that the LLVM code generated by this is pretty ugly (in particular, when the word to be overwritten is in the middle of the structure, it generates two `movmem` intrinsic calls), but the assembly generated by the LLVM compiler is quite compact.  The new `final-dump/mutate.wybe` test contains this type:

```
constructor make_position_tagged(int) | position( x:int, y:int, z:int)
```

The x86_64 assembler for procedures to update the x, y, and z, members, respectively, after checking the 1 tag on the input, look like this.  Note that when reading the existing position structure, all accesses are at an offset 1 less than the field in question, to compensate for the tag of 1 on the structure.  In each case, the  `incq    %rax` instruction near the end adds the 1 tag to the fresh structure.

**update_x**:

```
        movq    %rsi, %r14
        movl    $24, %edi
        callq   _wybe_malloc
        movq    7(%rbx), %rcx
        movq    15(%rbx), %rdx
        movq    %rcx, 8(%rax)
        movq    %rdx, 16(%rax)
        movq    %r14, (%rax)
        incq    %rax
        addq    $8, %rsp
        popq    %rbx
        popq    %r14
        retq
```

**update_y**:

```
        movq    %rsi, %r14
        movl    $24, %edi
        callq   _wybe_malloc
        movq    -1(%rbx), %rcx
        movq    %rcx, (%rax)
        movq    15(%rbx), %rcx
        movq    %rcx, 16(%rax)
        movq    %r14, 8(%rax)
        incq    %rax
        addq    $8, %rsp
        popq    %rbx
        popq    %r14
        retq
```

**update_z**:

```
        movq    %rsi, %r14
        movl    $24, %edi
        callq   _wybe_malloc
        movq    -1(%rbx), %rcx
        movq    7(%rbx), %rdx
        movq    %rcx, (%rax)
        movq    %rdx, 8(%rax)
        movq    %r14, 16(%rax)
        incq    %rax
        addq    $8, %rsp
        popq    %rbx
        popq    %r14
        retq
```


jimbxb:
I will review this. I just want to clean up the title/body so that GH recognises the issues will close :)

src/LLVM.hs
```diff
@@ -1024,7 +997,12 @@ writeLPVMCall "mutate" _ args pos = do
                     typeConvert struct struct2
                 ArgInt 0 _ -> do
                     logLLVM "lpvm mutate non-destructive case"
-                    duplicateStruct struct startOffset size struct2
+                    holeSize <- fromIntegral . (`div` byteBits) . typeRepSize
+                        <$> case (restIns,iRefs) of
+                        ([member],[]) -> argTypeRep member
+                        ([],[takeRef]) -> argTypeRep takeRef
+                        _ -> return $ Bits 0 -- error will be reported later
```
jimbxb: I'd rather have this error reported here. Maybe include the `shouldnt` in a where clause so we can reuse it here and where the error is currently reported?

src/LLVM.hs
```diff
@@ -1226,24 +1203,80 @@ llvmStoreValue ptr llVal = do
...
+                    Just (intConst realAfterHole, sz-realAfterHole))
+        _ -> do
+            -- Not a special case we handle:  just copy the whole struct
+            return (intConst 0, size, Nothing)
+    logLLVM $ "Duplicating struct with copy offset " ++ show copyOffset
+                ++ ", copy bytes " ++ show copyBytes
+                ++  ", extra " ++ show extra
+    -- Allocate memory
+    (writeDstCPtr,readDstCPtr) <- freshCPtrArgs
+    marshalledCCall mallocFn [] [size,writeDstCPtr] ["int","pointer"] Nothing
```
jimbxb: Should we reuse `heapAlloc` here?

jimbxb (changes requested):

jimbxb:
LGTM. Just some small nits.

---

There's a bunch of small, single line hunks that differ only in whitespace. I think that's caused by myself not properly trimming whitespace, and I assume your IDE does, causing these hunks to appear. 

Maybe we could get CI to run a formatter for us. But that's a job for another day.

PR #565, opened 2026-05-28
fix: correctly handle "ambiguity" of higher order and test bools

jimbxb:
The `:bool` constraint in the json parser sample bothered me.

The issue was, when the test `should_parse` was type checked, it could either be a call to a nil-adic higher-order function or a test. The "canonicalisation" step I added to avoid cases where the typing was the same, modulo type variables, didnt consider tests/HO calls correctly. This meant that `should_parse:()`

Because the HO typing from above was committed to first, the `pair(?should_parse, ?document)` call (or any of the other calls) never had its' chance to properly infer the `:bool` constraint.

### DIFF
--- a/samples/json.wybe
+++ b/samples/json.wybe
@@ -309,7 +309,7 @@ def print_list(ind:int, start:char, printer:{resource}(int, X), end:char, xs:lis
     pair(false, "{} {}")
 ]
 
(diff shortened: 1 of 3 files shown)


pschachte (approved):
LGTM

PR #566, opened 2026-05-28
feat: allow partial applications where all arguments are given

### LINKED ISSUE #562: Cannot pass a procedure as an argument by name
In the following example, it's possible to pass an anonymous procedure as an argument to `fetch_number`:
```
def fetch_number(f:(?int)):int = number where {
    f(?number)
}

?result = fetch_number({ ?@1 = 42 })
```
However, attempting the following will result in a compiler error: `Type error in call to proc fetch_number, argument 1`
```
def return_number(?result) {
    ?result = 42
}

def fetch_number(f:(?int)):int = number where {
    f(?number)
}

?result = fetch_number(return_number)
```
I’m not sure if this works as intended, but it feels intuitive to allow this behaviour.

jimbxb:
Closes #562


jimbxb:
Now the issue lies with `x = ?y` calls. They're being recognised as partial applications, and that ambiguity leads to these errors.

I think we can treat these as always being `move` instructions. I can't see why you'd want to define something like

```
def (a:X = ?b:Y) {
    foreign lpvm cast(x, ?y)
}
```
But I suppose there's nothing stopping you. It would just be awfully confusing!

---

Nevermind, if we want to support something like `` `=`(x):{test}(X,X) `` (being a partial application) then we can't treat all `x = ?y` as a move, as flattening would cause this partial application to appear as such.

jimbxb:
This is unfortunate. The testints test runs fine on my machine. It's just that the build times out in CI

jimbxb:
Diving a lot deeper into this, I see that there's too much ambiguity with the fmt and ,, calls in the string interpolation parts. Because the fmt call _could_ be a partial call (it's not, but the compiler doesn't knw), every one of the string interpolations leads to an extra recursive type check.

We could exclude any partial applications in the recursive type checks, but that seems like we'd just be pushing the issue down the line a bit. I'll have to think more on this, but I think the trick will be to order the type checks in such a way that they can fail faster, perhaps by enforcing`:string` constraints in all of the `fmt` calls in interpolations.

src/Scanner.hs
```diff
@@ -276,7 +276,8 @@ tokenisePart pos c cs
                     '!' -> singleCharTok c cs pos $ TokSymbol "!" pos
                     '@' -> singleCharTok c cs pos $ TokSymbol "@" pos
                     '\'' -> tokeniseChar pos cs
-                    '\"' -> tokeniseString DoubleQuote pos cs
+                    '\"' -> mapFst3 (([TokLBracket Paren pos] ++) . (++ [TokRBracket Paren pos]))
```
jimbxb, line 279: This is necessary for cases like `!print "$bar foo"`, as that was parsed as `!print fmt(bar) ,, " foo"`, and that caused issues due to the `!print` "binding" tighter than the `,,` when I added the `:wybe.string` constraints

pschachte: OK.  The approach I took with string interpolation was always a bit dodgy, a bit C preprocessor-like.  Throwing in extra parenthesis is perfectly reasonable.

src/Scanner.hs
```diff
@@ -373,22 +374,22 @@ tokeniseChar pos chars =
     where (front,back) = splitAt 2 chars
 
 -- |Tokenise a delimited string and tokenize the rest of the input..
-tokeniseString :: StringDelim -> SourcePos -> String -> PartialTokenisation
-tokeniseString delim pos cs =
+tokeniseString :: StringDelim -> Bool -> SourcePos -> String -> PartialTokenisation
```
pschachte, line 378: Please document the Boolean argument.

pschachte (changes requested):
All good except one requested bit of doc.

pschachte (approved):
Thank you!

PR #567, opened 2026-06-01
Improve type error for undefined variables/procs

pschachte:
Currently, undefined variables are reported as "proc bla unknown in ...".  It's not possible to distinguish between undefined variables and undefined niladic proc (ie, constants), but rather than assuming it's a proc, this change does not try to guess what it is, and just shows the name.


jimbxb (approved):

jimbxb:
LGTM. Just one more text exp

PR #569, opened 2026-06-05
fix: record extern LLVM fnptr in constants

jimbxb:
The test case exits with this on the a previous master build

```
Error detected during translating: 
llc: error: llc: /tmp/wybe-40074fe12aabd129/tmpMain.ll:8:78: error: use of undefined value '@unary_higher_order#.#anon#1<1>'
@"unary_higher_order#constant#0" = private unnamed_addr constant {ptr} { ptr @"unary_higher_order#.#anon#1<1>" }, align 8
```

I have also made the "env" param of closure procs explicit. This enabled further lowering of HO->FO calls, in some situations.

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -2352,7 +2352,7 @@ primImpurity (PrimHigher _ ArgVar{argVarType=HigherOrderType
     = return $ max impurity modimpurity
 primImpurity (PrimHigher _ (ArgConstRef structID _) impurity _) = do
     lookupConstInfo structID >>= \case
-        Just (StructInfo _ (FnPointerStructMember pspec:_)) ->
+        Just (ClosureInfo pspec _) ->
             max impurity . procImpurity <$> getProcDef pspec
         Just (VTableInfo _ (FnPointerStructMember pspec:t) _ _ _ _) ->
             max impurity . procImpurity <$> getProcDef pspec
@@ -3779,6 +3779,11 @@ data StructInfo
         structSize :: Int,          -- ^ The size of the struct in bytes
         structData :: [ConstValue]  -- ^ Contents of the struct, in memory order
         }
+    -- | A constant closure, with all params (needed or not)
+    | ClosureInfo {
+        closureProcSpec :: ProcSpec, -- ^ The closure proc
+        closureArgs :: [ConstValue]  -- ^ The closed args
+    }
     | VTableInfo {
         vtableSize :: Int,          -- ^ The size of the struct in bytes
         vtableData :: [ConstValue], -- ^ Contents of the struct, in memory order
@@ -3834,12 +3839,10 @@ constantValue _ = return Nothing
 closureStructId :: ProcSpec -> [PrimArg] -> Compiler (Maybe StructID)
 closureStructId pspec args = do
     mapM constantValue args >>= (\case
-        Just args' -> do
-          let sz = wordSizeBytes * (length args' + 1)
+        Just args' ->
           Just <$>
             recordConstStruct
-                (StructInfo sz
-                    $ FnPointerStructMember pspec : args')
+                (ClosureInfo pspec args')
                 (Just $ Closure pspec
                         (List.map (Unplaced . constValueExp) args'))
         _ -> return Nothing) . sequence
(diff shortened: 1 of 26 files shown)


src/AST.hs
```diff
@@ -3812,12 +3812,14 @@ data ArgFlowType = Ordinary        -- ^An argument/parameter as written by user
                                    -- ^An argument to pass a resource
                  | Free            -- ^An argument to be passed in the closure
                                    -- environment
+                 | ClosureEnv
```
pschachte: Please document this

test-cases/final-dump/box_unbox_ho.exp
```diff
@@ -517,7 +517,7 @@ proc #anon#1 > {inline} (1 calls)
     foreign llvm add(~anon#1#1##0:box_unbox, 2:box_unbox, ?anon#1#2##0:box_unbox) @box_unbox_ho:nn:nn
 proc #anon#1 > {inline} (1 calls)
 1: box_unbox_ho.#anon#1<1>
-#anon#1(anon#1#1##0 <{}; {}; {0}>, ?anon#1#2##1)<{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}>:
+#anon#1(@#env##0:opaque, anon#1#1##0 <{}; {}; {1}>, ?anon#1#2##1)<{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}>:
```
pschachte, line 520: Why is there an extra environment argument in this case?

test-cases/final-dump/unary_higher_order.exp
```diff
@@ -0,0 +1,90 @@
...
+
+define external fastcc void @"unary_higher_order#.<0>"() {
+  %"tmp#0##0" = tail call fastcc i64 @"unary_higher_order#.#anon#1<0>"()
+  call ccc void @print_int(i64 %"tmp#0##0")
+  ret void
+}
+
+define external fastcc i64 @"unary_higher_order#.#anon#1<0>"() {
+  ret i64 42
+}
```
pschachte: Nice!

test-cases/final-dump/unary_higher_order.exp
```diff
@@ -0,0 +1,90 @@
...
+  imports         : use wybe
+  resources       : 
+  procs           : 
+
+module top-level code > public {inline,semipure} (0 calls)
+0: unary_higher_order.<0>
+()<{<<wybe.io.io>>}; {<<wybe.io.io>>}; {}>:
+  AliasPairs: []
+  InterestingCallProperties: []
+    unary_higher_order.#anon#1<0>(?tmp#0##0:wybe.int) #0 @unary_higher_order:nn:nn
```
pschachte: This is awesome, but I don't understand why `unary_higher_order.#anon#1<0>` didn't get inlined here, when it's marked for inlining?

test-cases/final-dump/box_unbox_ho.exp
pschachte: I think I understand:  you're making this explicit so you can optimise it.  Is that the idea?

pschachte (changes requested):
One small change, and a question.

test-cases/final-dump/unary_higher_order.exp
jimbxb: I think it's because we only do one expansion/inlining pass. The call to fetch_number is expanded, and that lowers the HO call to a FO call, but we dont get the chance to inline the FO call

test-cases/final-dump/box_unbox_ho.exp
jimbxb: Yeah. It makes it mildly easier to align the arguments when optimising LPVM.

jimbxb: Okay. I want to do a bit more work on this.

I want to add explicit instructions to un-pack the closed values from the env as well. At the moment, there is still the assymetry of the env being present (and passed), but the free variables are not.

In order to do that, I need a special lpvm instruction. I'm calling it `unmarshall` for now. All it will do is take is the env param and the "index" of the free param of the closed variable in the non-closure variant of the proc, and return the closed value itself. 

I can't use an existing instruction is because the `unmarshall` instructions are special. They won't have a fixed offset to read from, as some free params may be unused in the non-closure proc. 

`unmarshall` could only be allowed in the pre-amble of a closure-proc when it comes to LLVM generation. We could also only inline a closure proc _if_ the env is known, too, which I believe would always be the case (we only get to the point we directly call the closure proc if we inline a HO call to a FO call!) 

The procs would look something like this:
```
def closure(#env, in, ?out) {
    foreign lpvm unmarshall(#env, 0, ?closed)
    non_closure(closed, in, ?out)
}

def non_closure(closed, in, ?out) {
    # body
}
```

src/AST.hs
jimbxb: Done

test-cases/final-dump/box_unbox_ho.exp
jimbxb: Actually, I dont know if this is possible. We still need _some_ of the information contained inside those "Free" params. We do still get that if the proc is a closure of some anonymous proc, but not when it is a closure like `` `+`(10)`` :(

I think I'll need to have both the explicit env and free params. I do still need to figure out how to factor in the `unmarshall` bit though. Let's see

jimbxb: Should be right now :) No need for `unmarshall`

jimbxb:
More merge conflicts 😭

PR #570, opened 2026-06-12
fix: trace aliases of constants through calls

jimbxb:
This fixes a bug I think has been around since #476 where constant structures are hoisted to global constants, but they are not treated correctly in the alias analysis, leading to cases where a global constant is mutated. 

I caught this when running one of the final dump tests, `box_unbox,wybe`, which currently segfaults due to this issue.

This has the unfortunate side effect of the `nbody.wybe` test case regressing back to 10001 allocations instead of 5001! This is because the value returned from `get_bodies` is a constant, and thus `advance` isnt able to be multi-spec'd.


pschachte (approved):
Thanks for catching and fixing that!  A bit unfortunate to lose all that specialisation of string.print, but this just makes improving aliasing analysis more important.

PR #572, opened 2026-06-18
fix: dont handle $ interpolations in c-strings

jimbxb:
Sorry, this one slipped my mind in #566. 

Added a test to be sure interpolations dont get tokenised when the flag is True.


pschachte (approved):
Fair enough.  In the long run, we may want to support interpolations in C strings using `strcat` or some such, but this is fine for now.

Issue #46, opened 2020-04-26
Allow a resource to be defined as a collection (ie, union) of other resources


pschachte:

pschachte:
Implemented in #573

PR #573, opened 2026-06-18
Compound resources

pschachte:
Allow resources to be defined as the union of other resources.  Using those resources then specifies that all the constituent resources are actually used.  This gives a more concise syntax for specifying use of a collection of resources.


pschachte:
Closes #46

src/Flatten.hs
```diff
@@ -67,23 +67,27 @@ flattenProcBody pd _ = do
                  $ List.map paramName
                  $ List.filter (flowsIn . paramFlow)
                  $ content <$> params
-        resources = List.map (resourceName . resourceFlowRes)
-                  $ procProtoResources proto
-        ProcDefSrc body = procImpln pd
+        resources = List.map resourceFlowRes $ procProtoResources proto
+
+    resources' <- List.map resourceName
+                . List.concatMap keys . catMaybes
```
jimbxb, line 73: `catMaybes` here will drop any unknown resources. Shouldnt we error?

src/Resources.hs
```diff
@@ -58,9 +58,83 @@ import Data.List.Extra (nubOrd, groupOn, groupSortOn)
...
+        expansion <- Map.lookup name . modResources <$> getLoadedModuleImpln mod
+        logResources $ " -> Known definition: " ++ show expansion
+        case expansion of
+            Just def -> do
+                logResources $ " -> Already expanded to " ++ show def
+                return def
+            Nothing -> do
+                let processing' = res:processing
+                (resSet, defPos) <-
+                    fromMaybe
```
jimbxb: Use trustFromJust here?

src/Flatten.hs
jimbxb: Ah, the expand step before this should avoid this. Can we make it error if there are any Nothings?

jimbxb (changes requested):
This program

```
resource foo = bar

pass
```
compiles with no error, but it ought to be erroneous as the `bar` resource is undefined.

pschachte:
OK, I think all issues have been addressed.

jimbxb:
Yes, looks good.

Though, a test case with a compound resource in a `use` block would be useful.

```
resource foo = bar, baz
resource bar:int = 1
resource baz:int

use foo in {
    ?baz = 1
    !use_baz
    !use_foo
    !use_bar
}

def use_foo use !foo, !io {
    !println("$bar, $baz")
    ?bar = 10
}
def use_bar use bar, !io {
    !println(bar)
}
def use_baz use baz, !io {
    !printin(baz)
}
```

jimbxb (approved):
LGTM

PR #574, opened 2026-06-23
feat: add trait declarations, implementations, and vtable dispatch

### LINKED ISSUE #538: Support bounded type quantification (type classes or interfaces or abstract types)
The goal is to make generics more useful, such as supporting a mapping type implemented as a binary tree where the key type is ordered, or supporting a sum operation of a list of any kind of number.

This can be implemented by making operations with generic arguments take an extra argument for each bounded generic argument, which would hold a vector of closures that implement the methods of the bounded generic type.  This is much like a vtable in OO language implementation.  We may want to make the restriction that the closures are simply function pointers, for efficiency.

I think the idea of an abstract type for this is attractive, as it would mostly unify the concepts of types and type classes.  An abstract type is simply a type that has some abstract procedures/functions.  That would make it easier to evolve a type into a type class with minimal code refactoring.

Abstract types would be declared with the keyword `abstract` and with no implementation.  Any type declared to `implement` the abstract type must have a definition for all its abstract procedures and functions.  For simplicity, these can simply be non-`abstract` procedures and functions with the right name, arity, and argument types and modes, defined in any module.  That means different modules can specify that a given type implements a given abstract type using different definitions for the methods; the selected methods will be those in scope where the call is made.

What distinguishes abstract types from type classes as in Haskell is that type classes are used by specifying that a type variable stands for an instance of a type class, and then that type variable is used in multiple places in the code to stand for the same type.  For example, in Haskell, `(==) :: Eq a => a -> a -> Bool`, which specifies that the types of the two arguments of `==` are the same.  In contrast, the natural expression of equality using an abstract `eq` type would be `a:eq = b:eq`.  One might expect that, if `int` and `float` both implement the `eq` type, this would mean that `=` could be applied to an `int` and a `float`, which is clearly nonsense.

I think the answer to this is to specify that, by default, all procedure/function arguments with the same abstract type must be instantiated to the same concrete type.  So `a:eq = b:eq` would require `a` and `b` to be the same type.  For the rare cases where you actually want to allow different arguments to have different types that implement the `eq` abstract type, we could allow a syntax like `operation(a:T1, b:T2) where T1<:eq, T2<:eq`.  Consider that when you declare `a:int = b:int`, you are saying not only that `a` and `b` are both `int`s, but also implicitly that `a` and `b` have the same type; this interpretation of specifying an abstract type as the type of variables just maintains that understanding.

Note that this issue is vaguely related to #255, #254, and #253, as all involve extensions to the Wybe type system.

wirelesseye:
Introduce trait-constrained type variables, abstract/default trait procedures, implementation validation, overload specificity, and LLVM vtable generation.

See [WYBE.md](https://github.com/wirelesseye/wybe/blob/feat/trait/WYBE.md#traits) for a description of this feature.

Closes #538


src/Clause.hs
```diff
@@ -401,6 +472,41 @@ compileParam allFlows startVars endVars procName idx param@(Param name ty flow f
     outIdx = if flowsIn flow then idx + 1 else idx
 
 
+compileVTableParam :: Int -> BoundedTypeVar -> Compiler [PrimParam]
+compileVTableParam index (varName, bound) = do
+    currMod <- getModuleSpec
+    boundTyp <- lookupType "compileVTableParam" Nothing bound
+    let boundMod = trustFromJust "compileVTableParam" $ typeModule boundTyp
+    return [PrimParam (PrimVarName "#vtable" index) (Representation CPointer)
```
jimbxb: Please extract `#vtable` as a constant, as other `#`-prefixed strings are

src/LLVM.hs
```diff
@@ -2336,6 +2390,18 @@ llvmLabelName :: String -> String
...
 
 
+-- | Make a suitable LLVM name for a vtable.
+llvmVTableName :: VTableSpec -> ModSpec -> Compiler String
+llvmVTableName (VTableSpec trait typ) mod = do
+    -- TODO: include ModSpec of the module where the vtable is defined
+    let typMod = trustFromJust "llvmVTableName" $ typeModule typ
+        traitMod = trustFromJust "llvmVTableName" $ typeModule trait
+    return $
+        "#vtable#" ++ showModSpec typMod
```
jimbxb: Extract this also

jimbxb: I think all the `#` should reference that constant, also.

jimbxb (commented):
This seems pretty cool. I'll try properly review it in some time

wirelesseye:
Convert to draft as the ABI differences between abstract and concrete procedures may cause issues on certain platforms (including arm64). Further work is needed to resolve this issue.

src/LLVM.hs
```diff
@@ -2336,6 +2390,18 @@ llvmLabelName :: String -> String
...
 
 
+-- | Make a suitable LLVM name for a vtable.
+llvmVTableName :: VTableSpec -> ModSpec -> Compiler String
+llvmVTableName (VTableSpec trait typ) mod = do
+    let typMod = trustFromJust "llvmVTableName" $ typeModule typ
+        traitMod = trustFromJust "llvmVTableName" $ typeModule trait
+    return $
+        vtableNamePrefix
+        ++ specialName (showModSpec typMod)
```
jimbxb: I'm not sure it's possible, but it might be worth mangling all of these, a la `mangleProcSpec`?

wirelesseye: So for now, a vtable name looks like this:
```
#vtable#path.to.trait#path.to.type#path.to.module
```
If we doing this in the `mangleProcSpec` style, should we use a different separator between trait, type and module? For example, use double hashes:
```
#vtable##path#.to#.trait##path#.to#.type##path#.to#.module
```
And one thing I forgot to handle is implementation type with type arguments. So for this `impl` declaration:
```
impl box(int) <: my_trait
```
The vtable name will be:
```
#vtable##path#.to#.my_trait##path#.to#.box(wybe#.int)##path#.to#.module
```
Does this look good to you?

jimbxb: Yeah, that _should_ be enough. I believe the only issue would be where there is a `.`, and that's what the mangler does

src/Clause.hs
```diff
@@ -411,6 +507,228 @@ compileParam allFlows startVars endVars procName idx param@(Param name ty flow f
...
+    compileABIParam (Param name ty flow ftype) =
+        [PrimParam (PrimVarName name 0) ty FlowIn ftype (ParamInfo False emptyGlobalFlows)
+            | flowsIn flow]
+        ++
+        [PrimParam (PrimVarName name (outNum flow)) ty FlowOut ftype (ParamInfo False emptyGlobalFlows)
+            | flowsOut flow]
+    outNum flow = if flowsIn flow then 1 else 0
+
+
+-- |Return the runtime type representations of the real ABI parameters, omitting
```
jimbxb: Don't we also have to be careful that all (real) parameters are marked as needed? I could see an implementation of some trait not needing some of the parameters (eg the `silent` one you had), and that would cause the ABI to be violated

jimbxb: There's probably also an issue if a wide (say `i128`) type is passed? This would be similar to the boxing issue for wide types when passed as a generic type

src/Config.hs
```diff
@@ -101,6 +106,16 @@ initProcName :: String
 initProcName = ""
 
 
+-- | Prefix used for generated vtable symbols.
+vtableNamePrefix :: String
+vtableNamePrefix = "#vtable"
```
jimbxb: Use `specialName "vtable"` here, etc :)

src/Clause.hs
wirelesseye: You're right, we should avoid eliminating parameters from adapter procs in Expansion. But then I feel like we need to generate an adapter for every vtable slot, regardless of whether the implementation proc's ABI seems compatible, to avoid potential parameter elimination.

jimbxb: Yeah. I think we should. That's how I implemented closures (via a "trampoline", and I think the pattern would be served well here too.

In most cases, the trampoline should be trivial enough that I'd hope it gets eliminated. But tbh I don't think I ever verified what the assembly looks like for them.

wirelesseye:
Hmm, this is weird. All tests passed on my machine, and I'm not sure what's caused the error in the CI build.

jimbxb:
That test has given me problems in the past because it takes a long time to type check. 

I haven't yet read the type checking changes, so can't pinpoint why/where it could be slowed, but generally speaking it'll be in a case where the type checker can't resolve the types and has to try to each of the possible types one-by-one.

pschachte:
Yes, the type checking algorithm is a bit inefficient.  Maybe for now just try quadrupling the timeout and see if that fixes the problem.  If so, let's move forward, and add as an issue the fact that the type checker is too inefficient.

src/AST.hs
```diff
@@ -1467,6 +1535,7 @@ data Module = Module {
   modIsType :: Bool,               -- ^Is this module a type, defined early
   modTypeRep :: Maybe TypeRepresentation, -- ^Type representation, when known
   modTypeSize :: Maybe Int,        -- ^The maximum size required to allocate an object of this type
+  modTrait :: Maybe (),            -- ^Is this module a trait, and its dependencies (placeholder)
```
pschachte: Why not just a Bool?

src/AST.hs
```diff
@@ -1752,6 +1843,12 @@ data ModuleImplementation = ModuleImplementation {
     modKnownResources :: Map Ident (Set ResourceSpec),
                                               -- ^Resources visible to this mod
     modKnownProcs:: Map Ident (Set ProcSpec), -- ^Procs visible to this module
+    modKnownTraitImpls :: Map TraitImplSpec (Placed (Maybe ModSpec)),
+                                              -- ^Trait impls visible to this mod
+                                              -- and the mods where they are defined
+                                              -- `Nothing` if defined in the current mod
```
pschachte: What is the `Placed` for?  Does the place make sense when it's `Nothing`?  If not, `Maybe (Placed ModSpec)` would make more sense.  Either way, please document.

src/AST.hs
```diff
@@ -1892,6 +1989,12 @@ data ImportSpec = ImportSpec {
     } deriving (Show, Generic)
 
 
+data ImportPhase =
```
pschachte, line 1998: Document what this is used for.

src/AST.hs
```diff
@@ -1971,23 +2080,51 @@ doImport mod (imports, _) = do
...
+-- |Publish the currently visible trait implementations in the module interface.
+publishTraitImpls :: Compiler ()
+publishTraitImpls = do
+    thisMod <- getModuleSpec
+    knownTraitImpls <- getModuleImplementationField modKnownTraitImpls
+    let traitImpls = Map.map (fromMaybe thisMod . content) knownTraitImpls
+    updateModInterface (\int -> int{ traitImpls=traitImpls })
+
+
+-- |Check if the second module is an ancestor of the first module
```
pschachte: Explain what *ancestor* means.  That's a new concept in Wybe.

src/AST.hs
```diff
@@ -2549,10 +2699,12 @@ emptyProcAnalysis = ProcAnalysis emptyDS Set.empty Map.empty
 
 isCompiled :: ProcImpln -> Bool
 isCompiled ProcDefPrim {} = True
-isCompiled (ProcDefSrc _) = False
+isCompiled ProcDefSrc{} = False
+isCompiled ProcDefAbstract{} = False
```
pschachte, line 2713: It's not clear that abstract procs should not be considered compiled.  After all, there's nothing to compile.  Please document why they are not considered compiled.

src/AST.hs
```diff
@@ -3569,20 +3749,43 @@ data PrimArg
      | ArgFloat Double TypeSpec                -- ^Constant floating point arg
      | ArgClosure ProcSpec [PrimArg] TypeSpec  -- ^Closure, with closed args
      | ArgGlobal GlobalInfo TypeSpec           -- ^Constant global reference
+     | ArgVTable (Either VTableSpec PrimArg) TypeSpec
```
pschachte: Why nest a `PrimArg`?  I would expect there would just be 2 possibilities:  a constant VTableSpec or a variable.  In the latter case, you probably just need the name.

src/AST.hs
```diff
@@ -3569,20 +3749,43 @@ data PrimArg
...
      | ArgUnneeded PrimFlow TypeSpec           -- ^Unneeded input or output
      | ArgUndef TypeSpec                       -- ^Undefined variable, used
                                                --  in failing cases
      deriving (Eq,Ord,Generic)
 
 
+-- |A trait specification.
+type TraitSpec = TypeSpec
+
+data TraitImplSpec =
```
pschachte, line 3776: Document

src/AST.hs
```diff
@@ -3569,20 +3749,43 @@ data PrimArg
...
+    TraitImplSpec {
+        implTrait :: TraitSpec,                -- ^The implmentation trait spec
+        implType  :: TypeSpec                  -- ^The implmentation type spec
+    }
+    deriving (Eq,Ord,Generic)
+
+type VTableSpec = TraitImplSpec
```
pschachte, line 3784: Document these two, too.

src/Parser.hs
```diff
@@ -657,6 +704,7 @@ data Associativity = LeftAssociative | NonAssociative | RightAssociative
 operatorAssociativity :: String -> (Int,Associativity)
 operatorAssociativity ":"  = (11, LeftAssociative)
 operatorAssociativity ":!" = (11, LeftAssociative)
+operatorAssociativity "<:"  = (11, RightAssociative)
```
pschachte, line 708: I don't think you should need this.  Associativity and precedence is determined from the last character of the name, which is a colon.  Try leaving this out and see if it still works.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+```
+
+This is equivalent to declaring a submodule named `speaker` whose first item is
+`trait`.
+
+An abstract procedure or function declaration has the same prototype syntax as
+a procedure or function declaration, but no body.  It must be in a trait module,
+and it must have exactly one type parameter constrained by that trait.  In the
+example above, `_` is an alias for the current trait, so the abstract function
+could instead be written as:
```
pschachte: You get a warning if you specify the current module by name, and that should apply when the current module is a trait.  So you would (or should) get a warning if you wrote `abstract speak(x:speaker): string`.  Instead say that the above declaration is equivalent to the below one.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+
+```
+abstract speak(x:speaker): string
+```
+
+Using a trait directly as a parameter type is a shorthand for a type variable
+constrained to that trait.  Writing the constraint explicitly is useful when
+there is more than one parameter with the trait bound. 
+
```
pschachte, line 1663: That's a bit unclear.  Say that if multiple parameters are given the trait type (`_`), then that implicitly specifies that they must all have the same type.  Then say that if it is desired to have two or more parameters constrained to implement the current trait, but not necessarily to have the same type, they must be written with the `<:` syntax, and specify different type variables.  An example would be helpful, `fmt2(x:T<:_, y:<T2<:_)` that calls `fmt` on both arguments without requiring them to have the same type.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+```
+def speak_twice(x:T<:speaker): string = "$(speak(x)) $(speak(x))"
+```
+
+Trait constraints can be combined.  Here `T` must implement both `speaker` and
+`named`:
+
+```
+def describe_speaker(x:T<:{speaker, named}): string = describe(x)
+```
```
pschachte, line 1709: Please find an example where the code really needs both traits.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+> `impl` *trait*
+
+For example, a `dog` type can implement `speaker` as follows:
+
+```
+# dog.wybe
+use speaker
+
+impl speaker
+
```
pschachte, line 1736: Can we `impl` multiple traits with a single `impl` declaration, like

```
impl trait1, trait2
```

If so, please document that.  If not, it would be nice to allow.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+```
+type printable trait {
+    abstract render(x:T<:_): string
+}
+
+impl int <: printable
+
+def render(x:int): string = "integer $x"
+```
```
pschachte, line 1782: Clarify that this assumes that `printable` requires only the method `render`.  Ideally, it would be better to stick to the same traits and methods you've used in earlier examples.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+
+The implementation procedures are resolved in the module containing the
+`impl` declaration.  If there are matching procedures imported from elsewhere,
+the local matching procedure is preferred.
```
pschachte: Clarify that the implementations *can* come from an imported module.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+
+A module becomes a trait by containing a `trait` declaration.  A trait cannot
+also declare a representation or constructors.  Its required operations are
+declared with `abstract`:
+
+```
+# speaker.wybe
+trait
+
+abstract speak(x:_): string
```
pschachte: The `speaker` trait and `speak` method are whimsical, but I think it would be more helpful to give an example people would actually use.  How about `fmt` as method (takes one argument and returns a string) and `formattable` as trait?  Or something else, if you like, but something that seems useful.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+
+pub constructor dog(name:string)
+
+pub def speak(x:_): string = "$(x^name) says woof"
+```
+
+Each abstract procedure or function in `speaker` must have exactly one matching
+concrete procedure or function for `dog`.  Its parameter flows, argument and
+result types, determinism, purity, and resource use must match the abstract
+declaration.
```
pschachte: This needs to be made clearer.  You should say that there must be a concrete (non-abstract) procedure or function that matches the signature of the abstract procedure/function, except that it has a parameter of the concrete type everywhere the abstract procedure/function has the trait type (or type variable with the trait as type bound).

test-cases/final-dump/trait_adapter_float.wybe
```diff
@@ -0,0 +1,9 @@
+type my_trait trait {
```
pschachte, line 4: Please put in a line or two of doc at the beginning of each test file saying what it tests for.

test-cases/final-dump/trait_error_unimplemented_type.exp
```diff
@@ -0,0 +1,3 @@
+Error detected during type checking of module(s) trait_error_unimplemented_type, trait_error_unimplemented_type.greeter
+final-dump/trait_error_unimplemented_type.wybe:9:10: Type error in call to proc greet_through_trait, argument 1
+final-dump/trait_error_unimplemented_type.wybe:9:10: Type error in partial application of proc greet_through_trait, argument 1
```
pschachte: That's not actually the right error message.  But that's OK for now; the type checker sometimes falls back to assuming a call is a partial application when there's no such HO procedure to call.  But it'd be a good idea to put a comment in the test case saying that the error message is wrong for now, and what it should say, so when we fix that, we'll know what it should be.

pschachte (changes requested):
This looks great!  I've pointed out a number of nits to fix, but otherwise this looks good to me.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+```
+type speaker trait {
+    abstract speak(x:_): string
+}
+```
+
+This is equivalent to declaring a submodule named `speaker` whose first item is
+`trait`.
+
+An abstract procedure or function declaration has the same prototype syntax as
```
jimbxb, line 1643: Please add some test cases/examples with proccs. Currently I see only funcs, yet the parser seems to support procs.

We should have a mix, to show that the abstract definition can be proc-like and the impl can be func-like, and vise versa.

WYBE.md
```diff
@@ -1610,6 +1611,143 @@ two input lists must be the same, and the result will be a list of the same
...
+impl speaker
+
+pub constructor dog(name:string)
+
+pub def speak(x:_): string = "$(x^name) says woof"
+```
+
+Each abstract procedure or function in `speaker` must have exactly one matching
+concrete procedure or function for `dog`.  Its parameter flows, argument and
+result types, determinism, purity, and resource use must match the abstract
```
jimbxb: We don't have any test cases for other detisms, purities, and resource usage. Please add them

src/AST.hs
wirelesseye: This is reserved for the trait dependency feature in the future. It will be changed to something like `Maybe (Set TraitSpec)`.

src/AST.hs
wirelesseye: Yes, for every trait impl, we always want to keep where the `impl` declaration is defined for error reporting, but we generate a vtable declaration or definition depending on whether it is defined in the current module. That's why I used `Maybe (Placed ModSpec)`: if it's from the current module then the value is `Nothing`. I'll update the comment to make it more clear.

src/AST.hs
wirelesseye: Because I feel like this makes expansion easier (in `expandArg'`)? I could be wrong though.

test-cases/final-dump/trait_error_unimplemented_type.exp
wirelesseye: What would the right error message look like? Do you mean it shouldn’t include line 3?

By the way, I believe that when we get compile-time duck typing for traits, this should throw the same error message as `trait_error_impl_missing`: 'Invalid implementation of trait ... for type ...: missing required proc ...'

src/Parser.hs
wirelesseye: `<:` does fall back based on its final colon, but I think the key distinction is the precedence. If we don't have this line, then this operator gets the default precedence 5, which won't allow us to parse a type like `T<:Trait` as a whole. I tried commenting out this line and it broke the parser, so I believe it's necessary for the current implementation.

src/AST.hs
wirelesseye: I'm not sure if this is verbose though…
```suggestion
    modKnownTraitImpls :: Map TraitImplSpec (Placed (Maybe ModSpec)),
      -- ^Trait impls visible to this module.  The `Placed` value always records
      -- where the impl declaration was defined for error reporting.
      -- `Maybe ModSpec` indicates whether the trait impl is external:
      --   `Nothing` means the impl is defined in the current module (local)
      --   `Just mod` means it is defined in another module (external)
```

src/AST.hs
wirelesseye: Done.

src/AST.hs
wirelesseye: Done.

src/AST.hs
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

test-cases/final-dump/trait_adapter_float.wybe
wirelesseye: Done.

src/AST.hs
wirelesseye: Done.

src/AST.hs
wirelesseye: Done.

WYBE.md
wirelesseye: I've rewritten the part. Hopefully it is clearer now.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Yeah, I've changed it to reuse the same traits in earlier examples.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

WYBE.md
wirelesseye: Done.

src/AST.hs
pschachte: OK, fair enough.  Please mention in the doc that this is a Maybe to prepare for storing type bounds.

src/AST.hs
pschachte: OK, that doc is clear enough.

src/AST.hs
pschachte: Have a look through the code that deals with this.  This representation means that an ArgVTable with the PrimArg case actually has two TypeSpecs, which could get out of sync, or it could have, say, an integer constant.  So you'll need to have a lot of messy sanity checking code.  If it doesn't make the code significantly messier, I think it would be better to simplify the PrimArg case to contain only the necessary info.

src/Parser.hs
pschachte: OK, fair enough.

test-cases/final-dump/trait_error_unimplemented_type.exp
pschachte: What's wrong about that error message is that the last line thinks the code is trying to be a partial application, when it's not.

test-cases/final-dump/trait_error_abstract_proc_purity.exp
```diff
@@ -0,0 +1,2 @@
+Error detected during type checking of module(s) trait_error_abstract_proc_purity, trait_error_abstract_proc_purity.observer
+final-dump/trait_error_abstract_proc_purity.wybe:7:1: Invalid implementation of trait trait_error_abstract_proc_purity.observer for type wybe.int: missing required proc observe
```
pschachte: This is likely to be confusing to the programmer.  Can you make the error message indicate that the provided method implementation doesn't match the abstract declaration.  Ideally, it would say that the implementation is impure where the spec says it should be semipure.  There is code in the type checker to report that kind of thing; maybe you could nest a TypeError inside the constructor for whatever you call your no-matching-implementation TypeError in cases where there is an implementation that matches name, arity, and types, but fails for other reasons, and print that to explain the problem.

pschachte (changes requested):
Almost all issues are resolved now.

pschachte (approved):
Thanks for all those fixes.

PR #576, opened 2026-06-29
feat: allow literal input expression as in-out

### LINKED ISSUE #554: Implied mode doesnt work properly for constants where an in-out is expected
This program ought to work
```
def foo(!x:int) {
    pass
}

foo(1)
```
by essentially "ignoring" the update.

Currently, this raises a `shouldnt`
```
Internal error: Cannot set type/flow of 1
  error, called at src/AST.hs:4686:17 in main:AST
```

There is one instance in the stdlib where I wish this would work
```
pub def c_string(s:_):c_string = str where {
    if { s = buffer(_, ?str) :: pass
       | else :: 
            ?len = length(s) + 1
            foreign lpvm alloc(len, ?str)
            foreign lpvm mutate(str, ?str, len - 1, true, len, 0, '\0')
            # Currently
            ?offset = 0
            pack(s, !str, len, !offset)
            # but I want it to be
            pack(s, !str, len, 0)
    }
}
```

### LINKED ISSUE #571: Input is implicitly passed where in-out is passed as out
This program 
```
def foo(!f) {
    !f + 1
}

?f = 1
foo(?f)
!println(f)
```
Prints `2`, but it ought to report an error.

Removing the `?f = 1` assignment leads to an error
```
def foo(!f) {
    !f + 1
}

# ?f = 1
foo(?f)
!println(f)
```
But only in clause generation
```
Error detected during generating low level code in testy
testy.wybe:6:6: Uninitialised variable 'f'
```

jimbxb:
Closes #554 
Closes #571 

Also adds some call stacks for `shouldnt`-like functions, for ease of debugging

### DIFF
--- a/samples/brainfuck.wybe
+++ b/samples/brainfuck.wybe
@@ -110,7 +110,7 @@ def run(instrs:list(instruction), !state:tape) use !io {
           | decr :: decr(!state)
           | left :: left(!state)
           | right :: right(!state)
-          | input :: !read(?state^current)
+          | input :: !read(!state^current)
           | output :: !print(state^current)
           | loop(?body) :: 
             do {
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -3399,6 +3399,7 @@ flattenedExpFlow (Var _ flow _)        = flow
 flattenedExpFlow (AnonParamVar _ flow) = flow
 flattenedExpFlow FailExpr              = ParamIn
 flattenedExpFlow (Typed exp _ _)       = flattenedExpFlow exp
+flattenedExpFlow (Global _)            = ParamIn
 flattenedExpFlow otherExp =
     shouldnt $ "Getting flow direction of unflattened exp " ++ show otherExp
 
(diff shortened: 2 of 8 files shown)


samples/brainfuck.wybe
```diff
@@ -110,7 +110,7 @@ def run(instrs:list(instruction), !state:tape) use !io {
           | decr :: decr(!state)
           | left :: left(!state)
           | right :: right(!state)
-          | input :: !read(?state^current)
+          | input :: !read(!state^current)
```
pschachte, line 113: Why change this?  The read is going to ignore the original value, so `?` is more precise than `!`.

pschachte (commented):
Aside from the query, this looks good.

samples/brainfuck.wybe
jimbxb: I changed it because it led to a compilation error, and I believe it should be one. The `?` really applies to `state`, not `state^current`, so the "original" value of the state isn't ignored.

After flattening in the current implementation, it looks like
```
!read(?tmp)
current(?state, tmp)
```
Mode-checking then turns this into
```
!read(?tmp)
current(!state, tmp)
```

That means that the `state` is implicitly passed in here. If this statement were in isolation (kinda like in the new test case), this could lead to a case where `state` is undefined before the call, and would be an error anyway.

After my fix, the formal `?` flow cannot match the actual `!` flow, as there's no "in" passed for the "in-out". I think that makes sense here

pschachte: Ah, yes, you're right.

pschachte (approved):

PR #578, opened 2026-07-16
fix: prevent duplicate vtables from being generated during LLVM emission

### LINKED ISSUE #577: Clashing vtable generation with traits in the standard library
I defined an `eq` trait in the standard library as `eq.wybe`:

```
trait

abstract {test} (a:_ = b:_)
abstract {test} (a:_ ~= b:_)
```

That compiles fine, but when I insert

```
use wybe.eq
impl eq
```

into `bool.wybe`, I get error messages compiling the standard library with `make`:

```
llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/c_array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/machine_word.ll:9:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/c_array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe/wybelibs/wybe/machine_word.ll:9:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^
```

wirelesseye:
Fix #577.

Before the fix, when generating a vtable for a trait implementation, the compiler only checks if the defining module of the vtable is an ancestor of the current module. However, it's possible that the ancestor modules have already *declared* the vtable, which results in duplicate vtable declarations in LLVM code.

This PR resolves this issue by preventing duplicate vtable entries from being recorded during LLVM emission.


pschachte (changes requested):
I tried this out with the library changes I mentioned in the issue, and find that this PR doesn't fix it.  I get these errors when trying to build the wybe library:

```
llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/c_array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/machine_word.ll:9:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/c_array.ll:8:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^

llc: error: llc: /Users/schachte/research/mine/wybe-test/wybelibs/wybe/machine_word.ll:9:1: error: redefinition of global '@#vtable##wybe#.eq##wybe#.bool##wybe#.bool#'
@"#vtable##wybe#.eq##wybe#.bool##wybe#.bool#" = external unnamed_addr constant {ptr, ptr}, align 8
^
```

wirelesseye:
@pschachte Hmm, that's weird. I can't reproduce the issue after the fix. I applied the same changes to `wybelibs`, and `make` compiled without errors. Perhaps try clearing the cache?

Fyi, I also tested with the following program. It doesn't import other modules except for the std library:
```
def {test} requires_eq(x: eq, y: eq) {
    x = y
}

!println(requires_eq(true, false))
!println(requires_eq(true, true))
```
This program compiles and prints the correct results:
```
false
true
```

jimbxb:
Maybe we should add some traits to the stdlib to try this out in CI? That would at least show if there is a discrepancy between Peter and your environments. We can always roll the commits back :)

src/Clause.hs
```diff
@@ -542,14 +543,17 @@ compileVTable vspec opmod = do
...
-        Nothing -> Just <$> recordConstStruct (VTableInfo sz values False vspec thisMod) Nothing
-
+    nestedIn <- getModuleImplementationField modNestedIn
+    generate <- case nestedIn of
+        Just nestedIn -> do
+            interface <- getModuleInterface `inModule` nestedIn
+            return $ not . Map.member vspec $ traitImpls interface
+        Nothing -> return True
+    if generate
+        then case opmod of
```
jimbxb: Would it be more clear to not have the case here?

```
            Just <$> recordConstStruct (VTableInfo sz values (isJust opmod) vspec (fromMaybe thisMod opmod)) Nothing
```

wirelesseye:
> Maybe we should add some traits to the stdlib to try this out in CI? That would at least show if there is a discrepancy between Peter and your environments. We can always roll the commits back :)

Sure we can, but changing the stdlib will break many test cases. If we were to add a trait implementation to `bool.wybe`, every module that imports `bool` would have an additional external vtable declaration. Is it okay to leave them as is for the test? Or should I update the `*.exp` files?

src/Clause.hs
wirelesseye: Right. I've made the change.

jimbxb:
Huh, I see. Is there a reason we need a module to declare the external vtables? Admittedly, I've not gotten a chance to read that part of the code, but I'd have thought that only `VTable` values that are used in the module would need to be declared?

Anyway, you could always remove the part of the code to dump those externals, just for now?

wirelesseye:
The tricky part is that if we don't generate external vtables at all, some test cases will be broken in a different way. Ideally, we should only generate an external vtable if it is used in the current module, as you said. I'll try to implement the check.

wirelesseye:
So the change passed CI. I can revert the last commit if we don't want to add `eq` to `wybelibs` at this stage.

pschachte:
Actually, I think we do want an `eq` trait in the standard library, and `bool` should surely implement it, so you might as well leave that in your branch for the PR.

pschachte:
Another point to consider:  it's not really whether the "parent module" has the same trait that's the issue, it's whether the declaration is going to appear in the same LLVM file.  It would be better to think of it that way when you're checking whether to generate the declaration.  The structure of the LLVM generation code collects *sets* of things that need to be declared from all the LPVM code that is going to be compiled into this LLVM file, and emits declarations once at the top of the file.  Using sets avoids repetition.  If you follow that strategy, you should avoid generating clashes.

wirelesseye:
I see. I’ll make that change.

wybelibs/wybe/eq.wybe
jimbxb, line 1: If we do keep this in, please don't forgwt to document it :)

pschachte: Here's something appropriate:
```
# purpose:  Standard trait describing types that support equality testing
# since  : 0.2
```

pschachte (approved):

PR #579, opened 2026-07-23
feat: change the naming convention for vtables

wirelesseye:
Currently, vtable names are verbose:
```
#vtable##trait_concrete_type_args#.printable##trait_concrete_type_args#.box(wybe#.float)##trait_concrete_type_args#
```
This PR changes the vtable naming convention to `<defining module>##vtable#<index>`. For example:
```
trait_concrete_type_args##vtable#0
```

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -23,7 +23,7 @@ module AST (
   determinismSeq, determinismProceding, determinismName, determinismCanFail,
   impurityName, impuritySeq, expectedImpurity,
   inliningName,
-  TraitSpec, TraitImplSpec(..), VTableSpec, TypeVarBound,
+  TraitSpec, TraitImplSpec(..), TypeVarBound,
   TypeProto(..), TypeModifiers(..), TypeSpec(..), typeVarSet, TypeVarName(..),
   genericType, higherOrderType, isHigherOrder,
   isResourcefulHigherOrder, isTraitType, typeModule,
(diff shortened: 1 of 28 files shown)


wirelesseye:
1. `structConstName` uses a single `#` as a separator and does not mangle `ModSpec`. Should we follow the same pattern?

	A regular struct constant vs a vtable constant:
	```
	trait_concrete_type_args#constant#0
	trait_concrete_type_args##vtable#0
	```

2. Should vtables have their own indices? For example:
	```
	"trait_concrete_type_args#constant#0"
	"trait_concrete_type_args#constant#1"
	"trait_concrete_type_args##vtable#0" 
	"trait_concrete_type_args##vtable#1" 
	"trait_concrete_type_args#constant#4"
	"trait_concrete_type_args#constant#5"
	```
	vs
	```
	"trait_concrete_type_args#constant#0"
	"trait_concrete_type_args#constant#1"
	"trait_concrete_type_args##vtable#2" 
	"trait_concrete_type_args##vtable#3" 
	"trait_concrete_type_args#constant#4"
	"trait_concrete_type_args#constant#5"
	```

jimbxb:
Since `constStructName` _doesnt_ mangle the modspsec, do you foresee an issue here where two constants of awfully named modules could ever collide? I suppose they can, but I'd also hope they get hoisted to the same constant before that could ever collide. Perhaps it's better to be safe than sorry here?

pschachte (approved):
I think this is OK as is.
We need to make name mangling consistent throughout:  procs, constants, vtables, everything.  I'll raise an issue to do that.

PR #580, opened 2026-07-23
Flesh out the eq trait for the std library

pschachte:
- Make all appropriate types implement the eq trait
 - Define `in` test for lists of eq type elements
 - Add a test of the latter for a couple of different eq types

### DIFF
--- a/wybelibs/wybe/range.wybe
+++ b/wybelibs/wybe/range.wybe
@@ -4,6 +4,9 @@
 pragma no_standard_library # Standard library can't depend on itself!
 
 use wybe.int, wybe.bool
+use wybe.eq
+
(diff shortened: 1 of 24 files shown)


wybelibs/wybe/range.wybe
```diff
@@ -4,6 +4,9 @@
 pragma no_standard_library # Standard library can't depend on itself!
 
 use wybe.int, wybe.bool
+use wybe.eq
+
+impl eq # type is canonical, so structural equality works
```
jimbxb, line 9: `in` would make sense here too

jimbxb (approved):
Nice. New optimisations! Surprised that was missed :(

PR #584, opened 2026-07-25
feat: allow trait types to have type parameters

wirelesseye:
This PR allows the declaration of trait types with type parameters:
```
type collection(T) trait {
    abstract insert(!collection:_(T), item:T)
    abstract contains(collection:_(T), item:T): bool
}

impl int_set <: collection(int)

def insert(!set:int_set, item:int) { ... }
def contains(set:int_set, item:int): bool = ...
```


pschachte (approved):
Nice!

PR #590, opened 2026-08-19
fix bug in decomposing a string with concat constructor

### LINKED ISSUE #589: using [|] to deconstruct a string with concat constructor causes a bus error
Example code:
```
for ?ch in "first" ,, "second" {
    !print(ch)
}
```

pschachte:
closes #589
Also add a regression execution test
The problem was that the code didn't respect the invariant that a `concat` constructor never has `empty` as either argument.

### DIFF
--- a/test-cases/execution/string_list_bug.exp
+++ b/test-cases/execution/string_list_bug.exp
@@ -1,0 +1,1 @@
+firstsecond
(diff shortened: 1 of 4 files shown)


jimbxb (approved):
Good catch!

PR #591, opened 2026-08-19
fix: add imported trait implementation conflict check

wirelesseye:
1. Add imported trait implementation conflict check. If imported modules contain the same trait implementation declaration but the implementation procedures are identical, it is not treated as an error.
2. Warn about duplicate trait implementations within a module.
3. Move support files for trait final dump test cases to a subdirectory.


src/AST.hs
```diff
@@ -1406,8 +1406,14 @@ addProcDef procDef = do
 
 addTraitImpl :: OptPos -> TraitImplSpec -> Maybe ModSpec -> Compiler ()
 addTraitImpl pos spec mod = do
-    updateImplementation (\imp -> imp {
-        modKnownTraitImpls = Map.insert spec (maybePlace mod pos) $ modKnownTraitImpls imp })
+    knownTraitImpls <- getModuleImplementationField modKnownTraitImpls
+    case Map.lookup spec knownTraitImpls of
+        Just _ | isNothing mod ->
```
pschachte, line 1411: Why `isNothing mod`?  Can't this happen when the module is explicit, too?

src/Normalise.hs
```diff
@@ -177,12 +177,41 @@ normaliseItem (PragmaDecl prag) =
...
+    knownTraitImpls' <- foldM insertNormalisedTraitImpl Map.empty normalised
     updateImplementation (\impl -> impl{ modKnownTraitImpls=knownTraitImpls' })
     publishTraitImpls
 
 
+-- |Insert a normalised trait implementation.  Distinct raw declarations can
+-- resolve to the same specification through qualification, aliases, or
+-- generic-variable canonicalisation.  Keep the earliest local declaration and
+-- warn at the later one.
+insertNormalisedTraitImpl :: Map TraitImplSpec (Placed (Maybe ModSpec))
```
pschachte, line 191: Please define and document a type for `Placed (Maybe ModSpec)`.  It's a somewhat surprising type.  It might also be worth defining a name for the `Map TraitImplSpec (Placed (Maybe ModSpec))` type if you use it much beyond this function.

src/Types.hs
```diff
@@ -625,6 +630,19 @@ typeErrorMessage (ReasonMultipleTraitImpl (TraitImplSpec trait typ) name pos) =
...
+    Message Error pos $
+        "Module " ++ showModSpec importingMod
+        ++ " imports conflicting implementations of trait " ++ show trait
+        ++ " for type " ++ show typ ++ ":"
+        ++ concatMap showImplementation implementations
+        ++ "\nDeclare a local implementation in module "
+        ++ showModSpec importingMod ++ " to override the imported implementations"
+  where
+    showImplementation (mod, procs) =
+        "\n    " ++ showModSpec mod ++ " uses {"
```
pschachte, line 644: `uses` usually refers to resources or module importation.  Can you think of another word to use here?

test-cases/final-dump/trait_basic.wybe
```diff
@@ -1,12 +1,16 @@
 # Verify trait dispatch and combined bounds across imported trait modules.
 
-use trait_basic_speaker, trait_basic_named, trait_basic_dog
+use trait_import_support.basic_speaker,
+    trait_import_support.basic_named,
+    trait_import_support.basic_dog
```
pschachte, line 6: It would be good to have at least one of these cases use the type name without module qualification.  Once the type is explicitly imported, the unmodule-qualified type name should work.

pschachte (changes requested):

src/Types.hs
jimbxb: Includes? Composed of?

src/AST.hs
wirelesseye: `isNothing mod` means the trait impl is declared in the current module. If `mod /= Nothing`, then it must from another module and therefore not considered a duplication (the local declaration overrides it). However, `addTraitImpl` right now is only called with `mod = Nothing`, and trait impls where `mod /= Nothing` are only generated during `handleModImports`, so maybe it's better to remove the `mod` parameter from `addTraitImpl`.

src/Normalise.hs
wirelesseye: I defined a type `KnownTraitImpl` for `Placed (Maybe ModSpec)`. As for `Map TraitImplSpec (Placed (Maybe ModSpec))` (now `Map TraitImplSpec KnownTraitImpl`), I feel it's fine, maybe? Because it's clear to see the purpose of the type now.

src/Types.hs
wirelesseye: I replaced `uses` with `provides`, because I think it reads more naturally as '(for this trait implementation, ) module A provides the following implementation procedures: …'

test-cases/final-dump/trait_basic.wybe
wirelesseye: Oh, that was a miss from a bulk replacement, because It didn't affect the correctness. I've changed other test cases to use the unqualified form as well.

src/AST.hs
pschachte: If you don't expect ever to need the `mod` parameter, it would be best to remove it.  If you think it will be need, then please put in a brief comment explaining that logic.

pschachte (approved):

PR #592, opened 2026-09-16
feat: support generic trait implementations with unbounded type variables

wirelesseye:
Support generic trait implementations with unbounded type variables:

```
type maybe_bool trait {
    abstract is_bool(x:_): bool
}

impl bool <: maybe_bool
impl T <: maybe_bool

def is_bool(x:bool): bool = true
def is_bool(x:T): bool = false

!println(maybe_bool.is_bool(true))
!println(maybe_bool.is_bool(42))
```

The outputs:
```
true
false
```


test-cases/final-dump/trait_error_impl_generic_proc.exp
```diff
@@ -0,0 +1,4 @@
+Error detected during type checking of module(s) trait_error_impl_generic_proc, trait_error_impl_generic_proc.maybe_int
+final-dump/trait_error_impl_generic_proc.wybe:8:1: Invalid implementation of trait trait_error_impl_generic_proc.maybe_int for type wybe.int: no matching implementation of proc is_int
+    candidate {test} trait_error_impl_generic_proc.is_int<0>(Type1) -> (Type1) does not match:
+        parameter 1 has type Type1, expected wybe.int
```
jimbxb: Any chance we can get this as `T`? I realise it's probably due to canonicalisation or unification somewhere, so might be a pain

wirelesseye: Done!

pschachte (approved):
All good.  Thanks.

PR #593, opened 2026-09-22
feat: support bounded generic trait implementation

wirelesseye:
Support bounded generic trait implementation:
```
type box(T) {
    pub box(value:T)
}

impl box(T<:formattable) <: named

def name(x:box(T<:formattable)): string = fmt(x^value)
```

## Limitation:

In the current implementation, there's no way to implement a bare type with trait bounds. Conceptually:
```
impl (T<:debug) <: printable
```
Because `(T)` is used for higher order types, I'm not sure if we want this syntax. But implementing it should be easy: only parser changes are required.


WYBE.md
```diff
@@ -1866,8 +1866,44 @@ specialised abstract procedures and functions.  In particular, a generic
...
+unbounded generic implementation.  Implementations with independent bounds
+may overlap without error, but using a type in their overlap is ambiguous if
+neither implementation is more specific.  A further implementation covering
+the intersection can resolve the ambiguity.  For example, given generic
+`pair(A,B)` and trait `named`:
+
+```
+impl pair(A<:formattable,B) <: named
+impl pair(A,B<:formattable) <: named
+impl pair(A<:formattable,B<:formattable) <: named
```
jimbxb: Spacing

src/AST.hs
```diff
@@ -5081,6 +5394,7 @@ instance Show StringVariant where
 instance Show GlobalInfo where
     show (GlobalResource res) = "<<" ++ show res ++ ">>"
     show (GlobalVariable res) = "@" ++ res
+    show (GlobalVTable res) = "global vtable " ++ show res
```
jimbxb: Is there a non-global cable more? Can we drop the prefix

wirelesseye: I wasn’t sure if the information would be helpful, but yes I can drop that

WYBE.md
```diff
@@ -1866,8 +1866,44 @@ specialised abstract procedures and functions.  In particular, a generic
...
+type box(T) {
+    pub box(value:T)
+}
+
+impl box(T<:formattable) <: named
+
+def name(x:box(T<:formattable)): string = fmt(x^value)
+```
+
+This implementation of `named` applies to `box(int)` if `int` implements
```
pschachte: In keeping with the idea that a trait name can be used as a type name, indicating "types that implement the specified trait", could the syntax

```
impl box(formattable) <: named
```

be supported as an abbreviation for `impl box(T<:formattable) <: named`?  Likewise

```
def name(x:box(formattable)): string = fmt(x^value)
```

Then the documentation should show this abbreviated version first, but then show the more long-winded version.

WYBE.md
pschachte: To my eyes, I'd like to see a space after the commas separating parameters, but not between the type variable, the `<:`,  and the trait name.

OTOH, I'd rather just see the first two written as

```
impl pair(formattable,B) <: named
impl pair(A,formattable) <: named
```

The third `impl` can't be written that way, because
```
impl pair(formattable,formattable) <: named
```

would specify a pair of two values of the same type, which must be formattable, rather than two values of possibly different formattable types.

test-cases/final-dump/trait_partial_impl_unsatisfied.wybe
```diff
@@ -0,0 +1,9 @@
+type hash trait { abstract hash(x:_): int }
```
pschachte: Please begin with a comment saying what this tests.  I can't immediately see what's wrong with it, except that it violates my expectations of what hashing and hash maps are.

test-cases/execution/trait_partial_impl_overlap_unused.wybe
```diff
@@ -0,0 +1,22 @@
+# Incomparable partial implementations are legal until their intersection is
```
pschachte: Why is this an execution test?  It doesn't seem to actually *do* anything related to traits; it just prints "ok". Either its behaviour should depend on traits, or it should just be a final dump test.

src/Types.hs
```diff
@@ -2876,13 +2812,15 @@ paramsMoreGeneral traitImpls general specific =
...
+        accepts gd sd generalPatterns specificPatterns =
+            typePatternsMoreGeneral gd sd satisfiesBound
+                generalPatterns specificPatterns
+        satisfiesBound bound ty = case resolveTraitImpl
+                (TraitImplSpec bound ty) traitImpls of
+            TraitImplResolved{} -> True
+            _ -> False
+    in accepts generalDict specificDict generalTypes specificTypes
+        && not (accepts specificDict generalDict specificTypes generalTypes)
   where
```
pschachte: It's a bit weird combining `where` and `let` clauses in the same function.  Please pick one.

pschachte (changes requested):

